PR #5, opened 2021-01-18
User: ListGroupMemberships

simongottschlag:
Changes:
- clients/user.go: Added `ListGroupMemberships` to [list user transitive memberOf](https://docs.microsoft.com/en-us/graph/api/user-list-transitivememberof?view=graph-rest-1.0&tabs=http)
- clients/user_test.go: Added tests for `ListGroupMemberships`
- auth/claims.go: Changed from `base64.StdEncoding.` to `base64.RawStdEncoding.` to fix error decoding access token. This is because JWTs are not padded which StdEncoding is expecting.

Any and all feedback, help, suggestions, edits etc is welcome!

### DIFF
--- a/auth/claims.go
+++ b/auth/claims.go
@@ -3,8 +3,9 @@ package auth
 import (
 	"encoding/base64"
 	"encoding/json"
-	"golang.org/x/oauth2"
 	"strings"
+
+	"golang.org/x/oauth2"
 )
 
 // Claims is used to unmarshall the claims from a JWT issued by the Microsoft Identity Platform.
--- a/auth/claims.go
+++ b/auth/claims.go
@@ -30,7 +30,7 @@ func ParseClaims(token *oauth2.Token) (claims Claims, err error) {
 		return
 	}
 	jwt := strings.Split(token.AccessToken, ".")
-	payload, err := base64.StdEncoding.DecodeString(jwt[1])
+	payload, err := base64.RawStdEncoding.DecodeString(jwt[1])
 	if err != nil {
 		return
 	}
--- a/clients/users.go
+++ b/clients/users.go
@@ -135,3 +135,31 @@ func (c *UsersClient) Delete(ctx context.Context, id string) (int, error) {
 	}
 	return status, nil
 }
+
+// GetMemberGroups returns a list of Groups the user is member of, optionally filtered using OData.
+func (c *UsersClient) GetMemberGroups(ctx context.Context, id string, filter string) (*[]models.Group, int, error) {
+	params := url.Values{}
+	if filter != "" {
+		params.Add("$filter", filter)
+	}
+	resp, status, err := c.BaseClient.Get(ctx, base.GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: base.Uri{
+			Entity:      fmt.Sprintf("/users/%s/transitiveMemberOf", id),
+			Params:      params,
+			HasTenantId: true,
+		},
+	})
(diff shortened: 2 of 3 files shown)


simongottschlag:
@manicminer is pagination supported by Golang / base or is it something that needs to be implemented? Not sure how many groups needs to be created to actually test it.

clients/users.go
```diff
@@ -135,3 +135,31 @@ func (c *UsersClient) Delete(ctx context.Context, id string) (int, error) {
 	}
 	return status, nil
 }
+
+// GetMemberGroups returns a list of Groups the user is member of, optionally filtered using OData.
+func (c *UsersClient) GetMemberGroups(ctx context.Context, id string, filter string) (*[]models.Group, int, error) {
```
manicminer, line 140: Could we rename this to `ListGroupMemberships`?

clients/users_test.go
```diff
@@ -100,3 +133,20 @@ func testUsersClient_Delete(t *testing.T, c UsersClientTest, id string) {
 		t.Fatalf("UsersClient.Delete(): invalid status: %d", status)
 	}
 }
+
+func testUsersClient_GetMemberGroups(t *testing.T, c UsersClientTest, id string) (groups *[]models.Group) {
```
manicminer, line 137: Likewise for the test, e.g. `testUsersClient_ListGroupMemberships`

manicminer (changes requested):
@simongottschlag Thanks for the PR, this is great! I just have one niggly request on the method name.

For pagination, there's no capability at present but I can look to add this quickly as it's essential and I'm sure to run into issues myself any time now without it.

clients/users.go
simongottschlag: No problem! I'm using the name that's in the graphrbac client right now, was thinking it would be easier to use the same.

I'll update tomorrow!

simongottschlag:
@manicminer I will update the name tomorrow. Used the same as they have in the graphrbac client because that's what I've been using.

I will create a new issue for pagination.

simongottschlag:
@manicminer I've updated the method names.

auth/claims.go
```diff
@@ -30,7 +31,7 @@ func ParseClaims(token *oauth2.Token) (claims Claims, err error) {
 		return
 	}
 	jwt := strings.Split(token.AccessToken, ".")
-	payload, err := base64.StdEncoding.DecodeString(jwt[1])
+	payload, err := base64.RawStdEncoding.DecodeString(jwt[1])
```
simongottschlag, line 34: @manicminer the reason RawStdEncoding is required is because JWTs are not padded, and StdEncoding expects padding.

https://www.rfc-editor.org/rfc/rfc7515.html#section-2

```
   Base64url Encoding
      Base64 encoding using the URL- and filename-safe character set
      defined in Section 5 of RFC 4648 [RFC4648], with all trailing '='
      characters omitted (as permitted by Section 3.2) and without the
      inclusion of any line breaks, whitespace, or other additional
      characters.  Note that the base64url encoding of the empty octet
      sequence is the empty string.  (See Appendix C for notes on
      implementing base64url encoding without padding.)
```

manicminer: Thanks this makes sense. I didn't hit any encoding issues in testing presumably because my JWT payloads were short (minimal set of claims).

manicminer (approved):

auth/claims.go
simongottschlag: @manicminer I think you have something in 1/4 to get padding or not. :+1:

Issue #6, opened 2021-01-18
Pagination


simongottschlag:
Hi!

When using List and methods like it, pagination will most likely be required when there are large amount of groups/users etc.

PR #7, opened 2021-01-19
OData pagination handling

manicminer:
Closes: #6

### DIFF
--- a/base/get.go
+++ b/base/get.go
@@ -1,9 +1,13 @@
 package base
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
 // GetHttpRequestInput configures a GET request.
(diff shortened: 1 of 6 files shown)


base/get.go
```diff
@@ -26,17 +31,81 @@ func (i GetHttpRequestInput) GetValidStatusFunc() ValidStatusFunc {
...
+		if odata.NextLink == nil || odata.Value == nil {
+			// No more pages, reassign response body and return
+			resp.Body = ioutil.NopCloser(bytes.NewBuffer(respBody))
+			return resp, status, nil
+		}
+
+		// Get the next page, recursively
+		nextInput := input
+		nextInput.rawUri = *odata.NextLink
+		nextResp, status, err := c.Get(ctx, nextInput)
```
simongottschlag, line 79: @manicminer my very basic understanding is that Golang doesn't have Tail Call Optimization which may (?) cause issues with recursive functions.

I don't know much about this, but been told by co-workers and heard it on a podcast.

Found this on SO: https://stackoverflow.com/questions/12102675/tail-call-optimization-in-go

Maybe a for loop should be used, or expose methods to the end users with something like Next() and let the end user page themselves?

PR #8, opened 2021-01-19
Service Principals: ListGroupMemberships

### LINKED ISSUE #1: Recursive groups for Service Principals
Hi!

Just found this project looking through the azuread terraform provider. Great work!

I've been building a reverse proxy that allows normal users and service principals to connect to any Kubernetes API and one part of this is extracting the groups they are members of.

I'm using a pure http client to get all the groups for a service principal, but after looking at this SDK it should fit quite well here instead.

Do you have any thoughts on adding this functionality to this SDK?

This is how I am doing it right now: https://github.com/XenitAB/azad-kube-proxy/blob/main/pkg/azure/service_principal.go

simongottschlag:
Closes #1 

Changes:
- clients/serviceprincipals.go: Added `ListGroupMemberships` to [list servicePrincipal transitive memberOf](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-list-transitivememberof?view=graph-rest-beta&tabs=http)
- clients/serviceprincipals_test.go: Added tests for `ListGroupMemberships`

### DIFF
--- a/clients/serviceprincipals.go
+++ b/clients/serviceprincipals.go
@@ -151,7 +151,6 @@ func (c *ServicePrincipalsClient) ListOwners(ctx context.Context, id string) (*[
 			Params:      url.Values{"$select": []string{"id"}},
 			HasTenantId: true,
 		},
-
 	})
 	if err != nil {
 		return nil, status, err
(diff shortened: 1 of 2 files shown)


simongottschlag:
@manicminer the next one :+1:

manicminer (approved):
Thanks @simongottschlag, LGTM 🚀

PR #11, opened 2021-01-25
Support MSI authentication, improve auth tests

manicminer:


manicminer:
Tested on a VM with managed service identity

PR #13, opened 2021-01-26
API rate limiting / retry logic

manicminer:
Adopt exponential backoff delay for requests returned with 424, 429 or 500 status

**For MS Graph:**
- start at 1 second, increase by 2^retries seconds up to 64 seconds
- If a "Retry-After" header is seen, parse for seconds and use that instead
- fail after 10 attempts

**For AAD Graph:**
- start at 5 seconds, increase by 2^retries seconds up to 64 seconds
- Parse any odata in the body and if an indicated backoff time is detected, use that instead
- fail after 10 attempts

### DIFF
--- a/base/aadgraph/base.go
+++ b/base/aadgraph/base.go
@@ -1,13 +1,13 @@
 package aadgraph
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


manicminer:
Rebased

PR #23, opened 2021-04-01
API Support for Conditional Access Policy

liammoat:
This PR will add API support for [conditionalAccessPolicy](https://docs.microsoft.com/en-gb/graph/api/resources/conditionalaccesspolicy?view=graph-rest-1.0) on the Microsoft Graph.


liammoat:
@manicminer, this is a draft PR to add API support for conditionalAccessPolicy. I'd like to add some tests for the ConditionalAccessPolicyClient. Can you advise if you have a specific tenant that the tests need to run against?

manicminer:
Hi @liammoat, thanks for this PR, this looks great!

For configuring tests, there is a helper at https://github.com/manicminer/hamilton/blob/main/clients/internal/testing.go which reads the auth configuration from environment variables, so it's up to the user to supply a tenant ID (in this case, for a P1 or P2 tenant) and client credentials.

If you can add some tests, I'll take another look and we can look to get this merged :)

clients/conditionalaccesspolicy.go
```diff
@@ -0,0 +1,147 @@
...
+func (c *ConditionalAccessPolicyClient) Update(ctx context.Context, conditionalAccessPolicy models.ConditionalAccessPolicy) (int, error) {
+	var status int
+	if conditionalAccessPolicy.ID == nil {
+		return status, errors.New("cannot update conditionalAccessPolicy with nil ID")
+	}
+
+	// This API does not handle PATCH on some properties
+	// i.e. cannot update values such as CreatedDateTime
+	conditionalAccessPolicy.CreatedDateTime = nil
+	conditionalAccessPolicy.ModifiedDateTime = nil
```
liammoat, line 114: @manicminer what are your thoughts on this?

The /identity/conditionalAccess/policies/%s endpoint does not allow these values to be passed to the PATCH request. 

I have test to protect against unwanted mutations. see line 76-78 in the test.go

manicminer: @liammoat I'm not sure this is needed when the `omitempty` flag is set in the struct tag? If you leave those fields as nil in the model when updating, they shouldn't be included in the body of the patch request.

If a user inadvertently specifies those when updating and receives an API error, I'd rather that were exposed than try to suppress it by nilling those fields.

clients/conditionalaccesspolicy_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/manicminer/hamilton/auth"
+	"github.com/manicminer/hamilton/clients"
+	"github.com/manicminer/hamilton/clients/internal"
+	"github.com/manicminer/hamilton/internal/utils"
+	"github.com/manicminer/hamilton/models"
+)
+
+var (
+	enterpriseAppId = os.Getenv("ENTERPRISE_APP_ID") // Enterprise Apps cannot be created using Graph. Therefore, to simplify testing pass the Enterprise App ID via an Env variable.
```
manicminer, line 16: Rather than requiring an ID be specified, could we use a service principal for Microsoft published app that's likely to be present in all tenants, e.g. Microsoft Graph? You can look those up by displayName or by appId.

manicminer (changes requested):
@liammoat Thanks for adding the tests. This mostly LGTM, I have a question about the enterprise application used for testing, and I replied to your query on the readonly field issue. 

Sorry for the inconvenience but I've done some refactoring that affects your PR - the `clients` and `models` packages are now a single package `msgraph`, with the models together alphabetically in `models.go`.

I also had to merge the same golangci-lint fix to the GH action so you'll probably need to revert that file.

If you can take a look at the above this should be good to merge. Thanks again!

clients/conditionalaccesspolicy.go
liammoat: Thanks @manicminer. Indeed `omitempty` removes it from the JSON payload if nil. I had an issue where the UPDATE test was "reusing" the object returned by the GET request test.

I have removed the code to nil these fields, and instead create a new object for the PATCH request with only the required fields `id` and `displayname`

This works well. Thanks for you feedback

clients/conditionalaccesspolicy_test.go
liammoat: Removed. I now pull an `appId` from the `PublishedApis` map. I can't use Graph, as it's not on the allowed list for first party applications in Conditional Access Policies. Instead, I opted for "AzureDevOps".

manicminer (approved):
Thanks for the changes @liammoat, this LGTM!

PR #30, opened 2021-04-13
Add support for directory role

michaljirman:
@manicminer thanks for merging the PR regarding the directory role templates. The next model which makes sense for me to add support for are the directory roles. 

This PR adds basic support for the following operation on the directory role object:
- listing activated directory roles within a tenant; only activated directory roles are available for listing/retrieving. A separate PR will follow adding support to activate the directory role template within a tenant. 
- retrieving a directory role by id
- listing, retrieving and updating members of a directory role

Notes: 
- API for the directory roles do not support PATCH so updating members in batches is not possible afaiu 
- Groups model was updated with an `IsAssignableToRole` attribute available only in the MS Graph Beta API. Groups client is already using beta API so there was no need for change. This attribute can be set only on the group's creation. It is also required to be set for directory role assignment. This operation is part of the tests.  
- Directory role assignments require at least P1 Azure AD Premium level


manicminer (approved):
@michaljirman Many thanks for this contribution. This LGTM and tests are working on my end 👍

PR #33, opened 2021-04-15
Update makefile

michaljirman:
Minor updates to makefile allowing devs to print out help sections. Also added fmt command for go files and enforcing race detection.


manicminer (approved):
Thanks @michaljirman, LGTM!

PR #37, opened 2021-04-30
add support for sending emails

michaljirman:
add support for sending emails; tiny

### DIFF
--- a/msgraph/me.go
+++ b/msgraph/me.go
@@ -69,3 +69,24 @@ func (c *MeClient) GetProfile(ctx context.Context) (*Me, int, error) {
 	}
 	return &me, status, nil
 }
+
+// SendMail sends message specified in the request body.
+func (c *MeClient) Sendmail(ctx context.Context, message MailMessage) (int, error) {
+	var status int
+	body, err := json.Marshal(message)
+	if err != nil {
+		return status, fmt.Errorf("json.Marshal(): %v", err)
+	}
+	_, status, _, err = c.BaseClient.Post(ctx, PostHttpRequestInput{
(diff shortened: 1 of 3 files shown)


manicminer:
Thanks @michaljirman, this is looking good! Could you add a test for each of those client methods?

michaljirman:
@manicminer I was thinking about it but it gets quite tricky to test. You need to use a special user with an ms365 licence for this. I am not sure how to create such a test user.

msgraph/me.go
```diff
@@ -69,3 +69,24 @@ func (c *MeClient) GetProfile(ctx context.Context) (*Me, int, error) {
 	}
 	return &me, status, nil
 }
+
+// SendMail sends message specified in the request body.
```
manicminer, line 73:
```suggestion
// SendMail sends message specified in the request body.
// TODO: Needs testing with an O365 user principal
```

msgraph/users.go
```diff
@@ -172,3 +172,24 @@ func (c *UsersClient) ListGroupMemberships(ctx context.Context, id string, filte
 	}
 	return &data.Groups, status, nil
 }
+
+// SendMail sends message specified in the request body.
```
manicminer, line 176:
```suggestion
// SendMail sends message specified in the request body.
// TODO: Needs testing with an O365 user principal
```

manicminer (approved):
@michaljirman That makes sense. Let's merge it without a test for now, and we can look at supporting tests with user principals (most likely using Azure CLI as mechanism) as there are going to be more resources with this requirement.

PR #38, opened 2021-05-01
Owned objects API for service principal

tsologub:
Add support for service principal "owned objects" API. Hidden, but quite handy functionality.

### DIFF
--- a/msgraph/serviceprincipals_test.go
+++ b/msgraph/serviceprincipals_test.go
@@ -40,6 +40,20 @@ func TestServicePrincipalsClient(t *testing.T) {
 		AppId:          app.AppId,
 		DisplayName:    utils.StringPtr(fmt.Sprintf("test-serviceprincipal-%s", c.randomString)),
 	})
+
+	appChild := testApplicationsClient_Create(t, a, msgraph.Application{
+		DisplayName: utils.StringPtr(fmt.Sprintf("test-serviceprincipal-child%s", a.randomString)),
+	})
+	spChild := testServicePrincipalsClient_Create(t, c, msgraph.ServicePrincipal{
+		AccountEnabled: utils.BoolPtr(true),
+		AppId:          appChild.AppId,
+		DisplayName:    utils.StringPtr(fmt.Sprintf("test-serviceprincipal-child%s", c.randomString)),
+	})
+
+	spChild.AppendOwner(string(c.client.BaseClient.Endpoint), string(c.client.BaseClient.ApiVersion), *sp.ID)
+	testServicePrincipalsClient_AddOwners(t, c, spChild)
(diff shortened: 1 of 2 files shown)


msgraph/serviceprincipals_test.go
```diff
@@ -157,6 +176,70 @@ func testServicePrincipalsClient_ListGroupMemberships(t *testing.T, c ServicePri
...
+		t.Fatalf("ServicePrincipalsClient.ListOwners(): %v", err)
+	}
+
+	if status < 200 || status >= 300 {
+		t.Fatalf("ServicePrincipalsClient.ListOwners(): invalid status: %d", status)
+	}
+
+	if len(*owners) != 1 {
+		t.Fatalf("ServicePrincipalsClient.ListOwners(): expected owners length 1. was: %d", len(*owners))
+	}
```
manicminer, line 220: I'm getting some failures here as it seems the SP inherits the application owners, which, in my testing tenant, gets autoassigned the calling SP as owner at creation time. This works around it for me, WDYT?

```suggestion
	ownersExpected := len(expected)

	if len(*owners) < ownersExpected {
		t.Fatalf("ServicePrincipalsClient.ListOwners(): expected at least 1 owner. has: %d", len(*owners))
	}

	var ownersFound int
	for _, e := range expected {
		for _, o := range *owners {
			if e == o {
				ownersFound++
				continue
			}
		}
	}
	if ownersFound < ownersExpected {
		t.Fatalf("ServicePrincipalsClient.ListOwners(): expected %d matching owners. found: %d", ownersExpected, ownersFound)
	}
```

manicminer (commented):
Thanks for this PR @tsologub! This mostly LGTM, I'm just getting some test failures for which I've proposed a workaround. Let me know what you think and once we have a solid test this should be good to merge :)

msgraph/serviceprincipals_test.go
tsologub, line 220:
```suggestion
	if len(*owners) == 0 {
		t.Fatalf("ServicePrincipalsClient.ListOwners(): expected at least 1 owner. has: %d", len(*owners))
	}
```
As I don't know the exact number of `expected` owners(your suggestion contains a non-existing `expected` variable), what would you say if we check as follows?

manicminer: My bad! I missed that when pasting. `expected` is an extra function arg - a string slice containing any expected owner UUIDs. The goal being to ensure at least a specific list of owners is correctly assigned.

manicminer (approved):
Sorry for the delay in re-reviewing @tsologub, this looks great and passes for me. Thanks again!

tsologub:
@manicminer no problem. When are you planning to make a `0.13.0` release?

manicminer:
@tsologub Should be sometime in the next day or so

Issue #36, opened 2021-04-28
Feature: Support servicePrincipals appRoleAssignments


Yanson:
We would like support for CRUD of service principals app role assignments.

https://docs.microsoft.com/en-us/graph/api/resources/approleassignment?view=graph-rest-1.0

This can then eventually be incorporated into the Terraform AzureAD provider to alleviate workarounds such as [this](https://github.com/hashicorp/terraform-provider-azuread/issues/164#issuecomment-725007993).

michaljirman:
@manicminer I will look into this issue. Feel free to assign it to me.

zbergus:
Good day, upon looking at the PR #39 i noticed the addition of (List/Remove/Assign)AppRoleAssignments in service principal which is awesome. Upon reading the graph documentation I noticed the Microsoft [suggests](https://docs.microsoft.com/en-us/graph/api/group-post-approleassignments?view=graph-rest-1.0&tabs=http) best practice for app role assignments is to use [AppRoleAssignedTo](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-list-approleassignedto?view=graph-rest-1.0&tabs=http). Is there a plan to add these 3 endpoints?

Using the existing code i extended the serviceprincipal struct for my current solution to add those 3 functionalities but I do not want to step on anyone's toes, here is what i user if it helps as a starting point:

```
// ListAppRoleAsssignedTo retrieves a list of appRoleAssignment that users, groups,
// or client service principals have been granted for the given resource service principal.
func (c *ServicePrincipalsClient) ListAppRoleAsssignedTo(ctx context.Context, id string) (*[]AppRoleAssignment, int, error) {
	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
		ValidStatusCodes: []int{http.StatusOK},
		Uri: Uri{
			Entity:      fmt.Sprintf("/servicePrincipals/%s/appRoleAssignedTo", id),
			Params:      url.Values{"$select": []string{"id"}},
			HasTenantId: true,
		},
	})
	if err != nil {
		return nil, status, fmt.Errorf("ServicePrincipalsClient.BaseClient.Get(): %v", err)
	}
	defer resp.Body.Close()
	respBody, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return nil, status, fmt.Errorf("ioutil.ReadAll(): %v", err)
	}
	var data struct {
		AppRoleAssignments []AppRoleAssignment `json:"value"`
[...]

PR #39, opened 2021-05-03
add support for app role assignments for users, groups and sp

michaljirman:
- Add support for app role assignments for users, groups and service principals
- Refactor existing app role assignments code to be used with all resource types (users, groups and service principals) while keeping resource type differences hidden
- Add app role assignments tests for users, groups and service principals

Supports app role assignments using the two available methods:
1. [appRoleAssignedTo relationship](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-list-approleassignedto?view=graph-rest-1.0&tabs=http) 

    client example
    ```
    msgraph.NewServicePrincipalsClient(...)
    ```

2. [appRoleAssignments relationship](https://docs.microsoft.com/en-us/graph/api/user-list-approleassignments?view=graph-rest-1.0&tabs=http)

    client examples
    ``` 
    msgraph.NewGroupsAppRoleAssignmentsClient(...)
    msgraph.NewUsersAppRoleAssignmentsClient(...)
    msgraph.NewServicePrincipalsAppRoleAssignmentsClient(...) 
    ```

Related to this [issue](https://github.com/manicminer/hamilton/issues/36)

### DIFF
--- a/msgraph/app_role_assignments.go
+++ b/msgraph/app_role_assignments.go
@@ -8,24 +8,50 @@ import (
 	"net/http"
 )
 
+type appRoleAssignmentsResourceType string
+
+const (
+	groupsAppRoleAssignmentsResource            appRoleAssignmentsResourceType = "groups"
+	usersAppRoleAssignmentsResource             appRoleAssignmentsResourceType = "users"
+	servicePrincipalsAppRoleAssignmentsResource appRoleAssignmentsResourceType = "servicePrincipals"
+)
+
 // AppRoleAssignmentsClient performs operations on AppRoleAssignments.
 type AppRoleAssignmentsClient struct {
-	BaseClient Client
+	BaseClient   Client
+	resourceType appRoleAssignmentsResourceType
 }
 
-// NewAppRoleAssignmentsClient returns a new AppRoleAssignmentsClient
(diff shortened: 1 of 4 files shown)


manicminer (commented):
Thanks for looking at this @michaljirman.

I was thinking about app role assignments recently and wondered if it would make more sense to implement all of the functionality as part of the GroupsClient, ServicePrincipalsClient and UsersClient structs rather than having separate clients. WDYT?

michaljirman:
@manicminer implementing method 2 within the GroupsClient, ServicePrincipalsClient and UsersClient will result in 3 near-identical methods in each of them and a lot of code duplication (appRoleAssignments relationship). We could still consider this refactor but it would be good to keep abstraction of these methods somewhere else and just call it from all of these clients. 

If we want to also keep method 1 we will additionally have extra 3 methods in the ServicePrincipalsClient (appRoleAssignedTo relationship)

michaljirman:
@manicminer we should probably get this reviewed/updated before I forget what it does :-)

manicminer:
@michaljirman Sorry for the delay in reviewing this, I've been focused on fixes and updates for the AzureAD Terraform provider the last 2 weeks. I'm planning to get this in along with #42 in the next few days.

manicminer (approved):
Thanks again @michaljirman, this LGTM  🚀

PR #41, opened 2021-05-06
add support for identity providers

michaljirman:
Add support for identity providers

- implements the [new endpoint](https://docs.microsoft.com/en-us/graph/api/resources/identityproviderbase?view=graph-rest-beta) in the beta API ([old endpoint for reference](https://docs.microsoft.com/en-us/graph/api/resources/identityprovider?view=graph-rest-beta))
- implements the List, Create, Get, Update, Delete and ListAvailableProviderTypes methods
- requires authentication with IdentityProvider.ReadWrite.All permission/scope

### DIFF
--- a/msgraph/identity_providers.go
+++ b/msgraph/identity_providers.go
@@ -1,0 +1,163 @@
+package msgraph
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// IdentityProvidersClient performs operations on IdentityProviders.
+type IdentityProvidersClient struct {
+	BaseClient Client
+}
+
+// NewIdentityProvidersClient returns a new IdentityProvidersClient
+func NewIdentityProvidersClient(tenantId string) *IdentityProvidersClient {
+	return &IdentityProvidersClient{
+		BaseClient: NewClient(VersionBeta, tenantId),
+	}
+}
+
+// List returns a list of IdentityProviders.
+func (c *IdentityProvidersClient) List(ctx context.Context) (*[]IdentityProvider, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      "/identity/identityProviders",
+			HasTenantId: true,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("IdentityProvidersClient.BaseClient.Get(): %v", err)
+	}
+	defer resp.Body.Close()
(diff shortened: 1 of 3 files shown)


msgraph/identity_providers.go
```diff
@@ -0,0 +1,163 @@
...
+	if err := json.Unmarshal(respBody, &provider); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+	return &provider, status, nil
+}
+
+// Update amends an existing IdentityProvider.
+func (c *IdentityProvidersClient) Update(ctx context.Context, provider IdentityProvider) (int, error) {
+	var status int
+	body, err := json.Marshal(provider)
```
manicminer, line 106: We should check the ID attribute to avoid a crash later on

```suggestion
	if provider.ID == nil {
		return status, errors.New("IdentityProvidersClient.Update(): cannot update identity provider with nil ID")
	}
	body, err := json.Marshal(provider)
```

msgraph/identity_providers_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/manicminer/hamilton/msgraph"
+)
+
+type IdentityProvidersClientTest struct {
+	connection   *test.Connection
+	client       *msgraph.IdentityProvidersClient
+	randomString string
+}
+
+func TestIdentityProvidersClientClient(t *testing.T) {
```
manicminer, line 18:
```suggestion
func TestIdentityProvidersClient(t *testing.T) {
```

msgraph/identity_providers_test.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		t.Fatalf("IdentityProvidersClient.Get(): %v", err)
+	}
+	if status < 200 || status >= 300 {
+		t.Fatalf("IdentityProvidersClient.Get(): invalid status: %d", status)
+	}
+	if provider == nil {
+		t.Fatal("IdentityProvidersClient.Get(): provider was nil")
+	}
+	return
```
manicminer, line 99: nil check here to avoid later crash (this needs adding in some other client tests too)

```suggestion
	if provider.ID == nil {
		t.Fatal("IdentityProvidersClient.Create(): provider.ID was nil")
	}
		return
```

manicminer (changes requested):
Hi @michaljirman, sorry for the delay in reviewing. This is looking great and tests are passing for me. If you can add some extra nil pointer checks this should be good to merge.

manicminer (approved):

manicminer:
@michaljirman Thanks again for this addition - I've committed a couple of suggestions and this is good to go.

michaljirman:
@manicminer sorry for the late reply was busy. Thanks for the additions and merge.

PR #48, opened 2021-05-25
List deleted API for apps, groups, and users

tsologub:
Add support for listing deleted applications, groups, and users. (i.e. basically [this](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-list?view=graph-rest-1.0) page)

Notes to the reviewer:
As 
> unexpected status 400 with OData error: Request_BadRequest: The service does not currently support writes of mail-enabled groups. Please ensure that the mail-enablement property is unset and the security-enablement property is set. 

is [not yet supported](https://docs.microsoft.com/answers/answers/371875/view.html) at the moment and it is a [long-desired feature](https://feedback.azure.com/forums/169401-azure-active-directory/suggestions/35083381-restore-azure-active-directory-security-groups) - I left the "Groups part" without a test case. I have tested this manually by creating+deleting O365 group and afterward executing `GroupsClient.ListDeleted()`.


tsologub:
Thanks for reviewing @manicminer.

> I've pushed a change to the ListDeleted methods so that they return slices of their respective struct type,

This totally makes sense. Now I wonder why I went with the slice of ids instead of the type. 🤔 

> and added a check to each of the tests to verify the recently deleted object is returned

Perfect. Even more robust tests.

PR #51, opened 2021-05-27
Get deleted API for apps, groups, and users

tsologub:
Add support for retrieving deleted applications, groups, and users. (i.e. basically [this](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-get?view=graph-rest-1.0&tabs=http) page)

Notes to the reviewer:
I wasn't quite sure how to approach this, as it is basically a code duplication. The HTTP request is the same (`/directory/deletedItems`), but different types will be returned in the end. I was thinking of introducing a `msgraph/directory.go` with `GetDeleted()`, but that didn't feel right. So I decided to keep this decoupled under separate clients, even if it is a code duplication somehow.

p.s. Groups without tests for the same (https://github.com/manicminer/hamilton/pull/48) reason.

### DIFF
--- a/msgraph/applications.go
+++ b/msgraph/applications.go
@@ -120,7 +120,7 @@ func (c *ApplicationsClient) GetDeleted(ctx context.Context, id string) (*Applic
 		},
 	})
 	if err != nil {
-		return nil, status, fmt.Errorf("ApplicationsClient.BaseClient.GetDeleted(): %v", err)
+		return nil, status, fmt.Errorf("ApplicationsClient.BaseClient.Get(): %v", err)
 	}
 	defer resp.Body.Close()
 	respBody, err := ioutil.ReadAll(resp.Body)
--- a/msgraph/applications.go
+++ b/msgraph/applications.go
@@ -109,6 +109,31 @@ func (c *ApplicationsClient) Get(ctx context.Context, id string) (*Application, 
 	return &application, status, nil
 }
 
+// GetDeleted retrieves a deleted Application manifest.
+// id is the object ID of the application.
+func (c *ApplicationsClient) GetDeleted(ctx context.Context, id string) (*Application, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/directory/deletedItems/%s", id),
+			HasTenantId: true,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("ApplicationsClient.BaseClient.GetDeleted(): %v", err)
+	}
+	defer resp.Body.Close()
+	respBody, err := ioutil.ReadAll(resp.Body)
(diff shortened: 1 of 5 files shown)


msgraph/applications.go
```diff
@@ -109,6 +109,31 @@ func (c *ApplicationsClient) Get(ctx context.Context, id string) (*Application,
...
+func (c *ApplicationsClient) GetDeleted(ctx context.Context, id string) (*Application, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/directory/deletedItems/%s", id),
+			HasTenantId: true,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("ApplicationsClient.BaseClient.GetDeleted(): %v", err)
```
manicminer, line 123:
```suggestion
		return nil, status, fmt.Errorf("ApplicationsClient.BaseClient.Get(): %v", err)
```

msgraph/groups.go
```diff
@@ -108,6 +108,30 @@ func (c *GroupsClient) Get(ctx context.Context, id string) (*Group, int, error)
...
+func (c *GroupsClient) GetDeleted(ctx context.Context, id string) (*Group, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/directory/deletedItems/%s", id),
+			HasTenantId: true,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("GroupsClient.BaseClient.GetDeleted(): %v", err)
```
manicminer, line 121:
```suggestion
		return nil, status, fmt.Errorf("GroupsClient.BaseClient.Get(): %v", err)
```

msgraph/users.go
```diff
@@ -106,6 +106,30 @@ func (c *UsersClient) Get(ctx context.Context, id string) (*User, int, error) {
...
+func (c *UsersClient) GetDeleted(ctx context.Context, id string) (*User, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/directory/deletedItems/%s", id),
+			HasTenantId: true,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("UsersClient.BaseClient.GetDeleted(): %v", err)
```
manicminer, line 119:
```suggestion
		return nil, status, fmt.Errorf("UsersClient.BaseClient.Get(): %v", err)
```

manicminer (approved):
Thanks again for this @tsologub. I think this approach is fine for now. I've been thinking about improving handling of generic directory objects but I feel this needs a few use cases first to get it right. Some code duplication is perfectly ok if it reads or works better.

PR #54, opened 2021-05-28
Delete permanently API for apps, groups, and users

tsologub:
Add support for deleting an already removed application/user/group permanently. ([doc](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-delete?view=graph-rest-1.0&tabs=http))


manicminer (approved):
Thanks @tsologub, this LGTM and the tests pass 👍

PR #55, opened 2021-05-28
NamedLocations: Set OData types in UpdateCountry / UpdateIP functions

kaovd:
The Odata type assignment is missed in the update functions which causes the requests to be ill formed and updates are not applied. See https://github.com/hashicorp/terraform-provider-azuread/pull/441


manicminer (approved):
Thanks @kaovd. I think we can probably make constants out of these API type strings, but this is OK for now 👍

PR #56, opened 2021-05-29
Named locations add generic get

alexwilcox9:
Creating an additional get function that works in a similar way to the list function. This will allow the terraform read function to more easily retrieve named locations on an import when the type is not known
See https://github.com/hashicorp/terraform-provider-azuread/pull/441 for reference


manicminer (approved):
Thanks @alexwilcox9, having a generic Get method makes sense. I'm seeing some consistency issues unrelated to your change but the test passes if I step through it slowly 👍

PR #57, opened 2021-06-02
Implementation for retrying requests

manicminer:
- Attempt to work around eventual consistency issues at the cost of some performance and potentially unnecessary requests.
- Request input structs can now contain a ConsistencyFailureFunc which determines whether a request should be retried on failure.
- ConsistencyFailureFuncs receive a copy of the response and the parsed OData with which to make such decisions.
- A canned RetryOn404ConsistencyFailureFunc can simply retry requests that receive a 404 response.
- This mitigates some eventual consistency issues, such as manipulating or referencing a newly created object, but doesn't work when listing objects.
- There's an example of a custom func in the ServicePrincipalsClient{}.Create() method which indicates retries when the referenced application isn't found


manicminer:
Test results with these mitigations:

Screenshot 2021-06-02 at 20 48 28

manicminer:
@romainDavaze @tsologub Do you have any thoughts on this approach? Is there a risk to replaying requests that I haven't realized?

In the end I figured it was probably better to enable this retry mechanism by default, but it can be disabled by setting `client.BaseClient.DisableRetries = true`.

romainDavaze:
This seems to be a good way to implement the retry mechanism.

To me the only risk is that the end-user might be a bit thrown off by the delay of function calls in some cases. I think it might be wise to add a note somewhere in the documentation explaining this behavior, why we do this and how it can be turned off.

This will introduce more delay in SDK calls which is not ideal, but I'd rather have this than something quicker that has a chance to fail.

tsologub:
I will gladly take a look on Monday (June 7). Currently, on a short vacation.

tsologub:
Great job! I like it, especially turning this feature on by default. 👍 Yeah, it can be slow, but a user can turn it off explicitly. 

Btw. I found other places where the retries can be useful as well. e.g. `application.RemovePassword()`, `application.GetDeleted()`, `directoryRoleTemplates.Get()`, and many more. Do you want to roll out the feature iteratively, and that's why left them untouched on purpose?

msgraph/client.go
```diff
@@ -126,6 +141,14 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
...
+	var reqBody []byte
+	if req.Body != nil {
+		reqBody, err = io.ReadAll(req.Body)
+		if err != nil {
+			return nil, status, nil, fmt.Errorf("reading request body: %v", err)
+		}
+	}
+
 	var attempts, backoff, multiplier int64
 	for attempts = 0; attempts < requestAttempts; attempts++ {
```
tsologub, line 153: It is slightly hard to understand for how long the retries will be performed. 
Correct me if I am wrong: 
- we have `10 attempts`.
- every time we have a consistency error, the `backoff` is set to `2`.
- in between attempts, we sleep for `2` seconds.

i.e., at most, we are retrying for ~20 seconds, right ?

***
Some time ago, I stumbled upon [this](https://youtu.be/AeLDtcpKc-U?t=418) video, where the author talks about `1-2 minute` for replication to propagate fully. 

We are daily provisioning applications and service principals, and I can say statistically it is about 2 minutes. i.e., we stopped receiving errors when we increased retries to 2 minutes mark.

I am afraid that ~20 seconds won't be enough. Or, we can start with something (`20 seconds`) and see how it goes.

manicminer:
Thanks all for the time to review  🙌

PR #58, opened 2021-06-09
Restore deleted API for apps, groups, and users

tsologub:
Add support for restoring deleted applications/users/groups.([doc](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-restore?view=graph-rest-1.0&tabs=http))


manicminer (approved):
Thanks @tsologub, this LGTM and tests passing 🚀

PR #61, opened 2021-06-17
Add Support For Sign-in,Directory Audit get/list actions and Create/List/Delete for Application Extensions

MarkDordoy:
This PR extends the models.go file to contain the required structs for SignInLogs, DirectoryAuditLogs and application extensions. 

It also has some basic tests which should be able to perform get and list on directory audit and signin logs as well as create, list and delete for application Extensions


MarkDordoy:
Tests past locally for me now. However the List SigninLogs can error out (this is dependent on how busy the tenant is and how many sign in logs there are to list).
I tired adding a top filter but it didnt seem to parse this correctly. eg:
```
func (c *SignInReportsClient) List(ctx context.Context, filter string, top string) (*[]SignInReport, int, error) {
	params := url.Values{}
	if filter != "" {
		params.Add("$filter", filter)
	}
	if top != "" {
		params.Add("$top", top)
	}
...
```
I think if we could get that working then i could ask the test to always just bring back top 1 so we get a speedy test. @manicminer  if you have any thoughts/idea's on this let me know.

MarkDordoy:
Hey @manicminer do i need to add anything else to get the approved and merged? 

Do you have any suggestions in terms of the tests to try and reduce the time and have the call just ask for the top 1?

manicminer:
Hey @MarkDordoy, sorry for the delay in reviewing.

The `$top` odata argument is its own query param rather than being a filter, so it needs to be added sort of like this:

```go
func (c *SignInReportsClient) List(ctx context.Context, filter string, top int) (*[]SignInReport, int, error) {
	params := url.Values{}
	if filter != "" {
		params.Add("$filter", filter)
	}
	if top > 0 {
		params.Add("$top", fmt.Sprintf("%d", top))
	}
```

This isn't very wieldy though - there are lots of requests that support $top and I don't think it'd be nice to add another argument to a load of client methods. I'm not entirely happy about the `filter string` pattern to be honest - this feels like a good time to add more robust odata query handling across the board.

That aside, this PR looks great. I made a few small changes locally - I renamed the `AuditLog` model to `DirectoryAudit` to match the API model and added a couple type aliases for user-specified enums. I also took the chance to sort the models. The tests pass for me and with these changes this will be good to merge.

Could you [enable upstream contributions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) for this PR and I'll push my changes, then we can get this merged? Many thanks for the contribution! :)

MarkDordoy:
Hey @manicminer I've selected the "Allow edits by maintainers" button in this PR, i think thats what you need to be able to push commits. Let me know if it doesnt work

manicminer:
Thanks, that did it - I also had your branch name wrong 😅

PR #62, opened 2021-06-19
Support TermsOfUse Agreements 

kaovd:
Introduces ability to configure TermsOfUse Agreements as documented in relevant 1.0 API Spec as a subsection of conditional access - https://docs.microsoft.com/en-us/graph/api/agreement-post-agreements?view=graph-rest-1.0&tabs=http


kaovd:
Need to quickly rewrite this, I wrote it according to what I was seeing in the portal API respones but I've learned the hard way that what Graph Explorer comes back with vs Azure AD Portal is totally different -_-

kaovd:
Was wondering why I was getting 403's didnt realise az login --service-principal was being used.

```
=== RUN   TestTermsOfUseAgreementClient
--- PASS: TestTermsOfUseAgreementClient (31.92s)
PASS
ok      github.com/manicminer/hamilton/msgraph  31.927s
```

manicminer:
Hi @kaovd, many thanks for this PR. I've rebased and squashed your changes as models.go was recently touched throughout.

This mostly LGTM, however I noticed that the `files` collection is deprecated in favor of `file` and `file/localizations`, I think we should rather support the newer properties? Also can you test to see if files can be updated or if this requires a separate request to a relationship endpoint? Thanks!

kaovd:
@manicminer so Portal behaviour is that you can update agreements and add new language versions, but you cannot change existing language versions or delete them. Should be handled in AzureAD Provider for a force replacement condition. Will test, also will take a look at updating

kaovd:
I'm actually struggling to find a properly whole / concise example for this new format so while I would try and use the non-depcreated option and im not really sure where the full spec is.

PR #68, opened 2021-06-24
Add Support For Schema Extensions

MarkDordoy:
This PR creates the models, data types and functions to perform CRUD operations on schema Extensions.

https://docs.microsoft.com/en-us/graph/api/resources/schemaextension?view=graph-rest-1.0

### DIFF
--- a/msgraph/valuetypes.go
+++ b/msgraph/valuetypes.go
@@ -52,14 +52,15 @@ const (
 type ExtensionSchemaTargetType string
 
 const (
-	ExtensionSchemaTargetTypeContact      ExtensionSchemaTargetType = "Contact"
-	ExtensionSchemaTargetTypeDevice       ExtensionSchemaTargetType = "Device"
-	ExtensionSchemaTargetTypeEvent        ExtensionSchemaTargetType = "Event"
-	ExtensionSchemaTargetTypePost         ExtensionSchemaTargetType = "Post"
-	ExtensionSchemaTargetTypeGroup        ExtensionSchemaTargetType = "Group"
-	ExtensionSchemaTargetTypeMessage      ExtensionSchemaTargetType = "Message"
-	ExtensionSchemaTargetTypeOrganization ExtensionSchemaTargetType = "Organization"
-	ExtensionSchemaTargetTypeUser         ExtensionSchemaTargetType = "User"
+	ExtensionSchemaTargetTypeAdministrativeUnit ExtensionSchemaTargetType = "AdministrativeUnit"
+	ExtensionSchemaTargetTypeContact            ExtensionSchemaTargetType = "Contact"
+	ExtensionSchemaTargetTypeDevice             ExtensionSchemaTargetType = "Device"
+	ExtensionSchemaTargetTypeEvent              ExtensionSchemaTargetType = "Event"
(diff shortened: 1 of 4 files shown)


msgraph/valuetypes.go
```diff
@@ -49,6 +49,29 @@ const (
...
+
+const (
+	ExtensionSchemaTargetTypeContact      ExtensionSchemaTargetType = "Contact"
+	ExtensionSchemaTargetTypeDevice       ExtensionSchemaTargetType = "Device"
+	ExtensionSchemaTargetTypeEvent        ExtensionSchemaTargetType = "Event"
+	ExtensionSchemaTargetTypePost         ExtensionSchemaTargetType = "Post"
+	ExtensionSchemaTargetTypeGroup        ExtensionSchemaTargetType = "Group"
+	ExtensionSchemaTargetTypeMessage      ExtensionSchemaTargetType = "Message"
+	ExtensionSchemaTargetTypeOrganization ExtensionSchemaTargetType = "Organization"
+	ExtensionSchemaTargetTypeUser         ExtensionSchemaTargetType = "User"
```
manicminer, line 62:
```suggestion
	ExtensionSchemaTargetTypeAdministrativeUnit ExtensionSchemaTargetType = "AdministrativeUnit"
	ExtensionSchemaTargetTypeContact            ExtensionSchemaTargetType = "Contact"
	ExtensionSchemaTargetTypeDevice             ExtensionSchemaTargetType = "Device"
	ExtensionSchemaTargetTypeEvent              ExtensionSchemaTargetType = "Event"
	ExtensionSchemaTargetTypePost               ExtensionSchemaTargetType = "Post"
	ExtensionSchemaTargetTypeGroup              ExtensionSchemaTargetType = "Group"
	ExtensionSchemaTargetTypeMessage            ExtensionSchemaTargetType = "Message"
	ExtensionSchemaTargetTypeOrganization       ExtensionSchemaTargetType = "Organization"
	ExtensionSchemaTargetTypeUser               ExtensionSchemaTargetType = "User"
```

manicminer (commented):
@MarkDordoy thanks for the PR! I rebased and force-pushed your branch to make GitHub use the correct workflow configs, but this mostly LGTM. I just noticed that `AdministrativeUnit` is missing from the `ExtensionSchemaTargetType` constants?

msgraph/valuetypes.go
MarkDordoy: Hey @manicminer sorry for slow response, i was away. 

So i dont believe `AdministrativeUnit` is supported at least for now by Schema extensions. I may be wrong, im basing my understanding of this from this doc: https://docs.microsoft.com/en-us/graph/api/resources/schemaextension?view=graph-rest-1.0 which details:
```Set of Microsoft Graph types (that can support extensions) that the schema extension can be applied to. Select from contact, device, event, group, message, organization, post, or user.```

manicminer: Ah right, we're actually using the beta endpoint in this client which is where I got the list of values from (https://docs.microsoft.com/en-us/graph/api/resources/schemaextension?view=graph-rest-beta). We could add the value, or switch to the v1.0 endpoint?

manicminer (approved):
Thanks @MarkDordoy, this LGTM 🚀

PR #72, opened 2021-06-28
Extend User model - Add SignInActivity

MarkDordoy:
Add SignInActivity to User struct.

This is only available currently via the beta API.

Returned only on $select. Supports $filter (eq, ne, NOT, ge, le) but, not with any other filterable properties. Note: Details for this property require an Azure AD Premium P1/P2 license and the AuditLog.Read.All permission.


MarkDordoy:
From testing this works fine with corporate tenants but does not appear in the Gov cloud tenants yet

manicminer (approved):
Thanks @MarkDordoy, this LGTM 👍

PR #78, opened 2021-07-16
Ability to inject middleware functions to manipulate requests and responses

manicminer:
- Middleware functions can be stacked and provided for both Request and Response handling
- Replace the request retry logic with https://github.com/hashicorp/go-retryablehttp
- Export `msgraph.Client{}.HttpClient` so callers can additionally supply their own http.Client

### DIFF
--- a/msgraph/client.go
+++ b/msgraph/client.go
@@ -34,6 +34,12 @@ const (
 // ConsistencyFailureFunc is a function that determines whether an HTTP request has failed due to eventual consistency and should be retried
 type ConsistencyFailureFunc func(*http.Response, *odata.OData) bool
 
+// RequestMiddleware can manipulate or log a request before it is sent
+type RequestMiddleware func(*http.Request) (*http.Request, error)
+
+// ResponseMiddleware can manipulate or log a response before it is parsed and returned
+type ResponseMiddleware func(*http.Request, *http.Response) (*http.Response, error)
+
 // RetryOn404ConsistencyFailureFunc can be used to retry a request when a 404 response is received
 func RetryOn404ConsistencyFailureFunc(resp *http.Response, _ *odata.OData) bool {
 	return resp.StatusCode == http.StatusNotFound
@@ -77,6 +83,12 @@ type Client struct {
 	// DisableRetries prevents the client from reattempting failed requests (which it does to work around eventual consistency issues).
 	// This does not impact handling of retries related to rate limiting, which are always performed.
 	DisableRetries bool
+
+	// RequestMiddlewares is a slice of functions that are called in order before a request is sent
+	RequestMiddlewares []RequestMiddleware
+
+	// ResponseMiddlewares is a slice of functions that are called in order before a response is parsed and returned
+	ResponseMiddlewares []ResponseMiddleware
 
 	httpClient *http.Client
 }
@@ -164,9 +176,26 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
 		}
 
 		req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
+
+		for _, m := range c.RequestMiddlewares {
+			r, err := m(req)
+			if err != nil {
+				return nil, status, nil, err
+			}
+			req = r
+		}
+
 		resp, err = c.httpClient.Do(req)
 		if err != nil {
 			return nil, status, nil, err
+		}
+
+		for _, m := range c.ResponseMiddlewares {
+			r, err := m(req, resp)
+			if err != nil {
+				return nil, status, nil, err
+			}
+			resp = r
 		}
 
 		o, err = odata.FromResponse(resp)
--- a/msgraph/client.go
+++ b/msgraph/client.go
@@ -151,7 +151,7 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
 	}
 
 	c.retryableClient.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
-		if !c.DisableRetries {
+		if resp != nil && !c.DisableRetries {
 			if resp.StatusCode == http.StatusFailedDependency {
 				return true, nil
 			}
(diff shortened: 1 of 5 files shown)


msgraph/client.go
```diff
@@ -85,22 +78,25 @@ type Client struct {
 	DisableRetries bool
 
 	// RequestMiddlewares is a slice of functions that are called in order before a request is sent
-	RequestMiddlewares []RequestMiddleware
+	RequestMiddlewares *[]RequestMiddleware
```
tombuildsstuff, line 81: I think this wants to become:

```suggestion
	RequestMiddlewares []*RequestMiddleware
```

which'd mean the middleware functions themselves would be pointers?

msgraph/client.go
```diff
@@ -85,22 +78,25 @@ type Client struct {
...
 
 	// ResponseMiddlewares is a slice of functions that are called in order before a response is parsed and returned
-	ResponseMiddlewares []ResponseMiddleware
+	ResponseMiddlewares *[]ResponseMiddleware
```
tombuildsstuff, line 84: (same here)?

msgraph/client.go
```diff
@@ -161,100 +149,73 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
...
 
-		if !c.DisableRetries {
-			f := input.GetConsistencyFailureFunc()
-			if f != nil && f(resp, o) {
-				// eventual consistency, just try again
-				if attempts < requestAttemptsForConsistency {
-					continue
-				}
-			}
+	status = resp.StatusCode
```
tombuildsstuff, line 204: what if `resp` is nil here (for example a dropped connection?)?

msgraph/client.go
```diff
@@ -161,100 +149,73 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
...
-					return nil, status, o, fmt.Errorf("unexpected status %d, could not read response body", resp.StatusCode)
-				}
-				errText = fmt.Sprintf("response: %s", respBody)
+		var errText string
+		switch {
+		case o != nil && o.Error != nil && o.Error.String() != "":
+			errText = fmt.Sprintf("OData error: %s", o.Error)
+		default:
+			defer resp.Body.Close()
+			respBody, err := io.ReadAll(resp.Body)
```
tombuildsstuff, line 217: these two lines assume there's a body, but we're not checking that it has any length here?

msgraph/client.go
```diff
@@ -161,100 +149,73 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
 		}
 	}
 
-	var attempts, backoff, multiplier int64
-	for attempts = 0; attempts < requestAttemptsForRateLimiting; attempts++ {
-		// sleep after the previous failed attempt
-		if attempts > 0 {
-			time.Sleep(time.Duration(backoff))
-		}
+	c.retryableClient.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
```
tombuildsstuff, line 153: because this function sits on the instance of the object `(Client)` and not `(*Client`) this change to retryableClient won't be retained, which seems fine but just double-checking

msgraph/client.go
```diff
@@ -78,17 +77,27 @@ type Client struct {
...
-		UserAgent:  "Hamilton (Go-http-client/1.1)",
-		httpClient: &http.Client{},
+		Endpoint:        environments.MsGraphGlobal.Endpoint,
+		ApiVersion:      apiVersion,
+		TenantId:        tenantId,
+		UserAgent:       "Hamilton (Go-http-client/1.1)",
+		HttpClient:      r.StandardClient(),
+		retryableClient: r,
 	}
 }
```
tombuildsstuff, line 102: unrelated to this PR, but a though that it'd be convenient to have a `NewClientWithEndpoint` constructor that this calls with `environments.MsGraphGlobal.Endpoint`?

tombuildsstuff (approved):
A few minor comments but this otherwise LGTM 👍

msgraph/client.go
```diff
@@ -149,83 +150,77 @@ func (c Client) performRequest(req *http.Request, input HttpRequestInput) (*http
...
-		if cap := int64(retryBackoffDelayCap); backoff > cap {
-			backoff = cap
-		}
-
-		req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
-		resp, err = c.httpClient.Do(req)
-		if err != nil {
-			return nil, status, nil, err
-		}
+	c.retryableClient.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
```
adamconnelly, line 153: Just a thought - is it possible / worthwhile to export this default retry function somehow? I'm just thinking, what if I want to pass in a custom HTTP client, but I still want to use the same retry behaviour? That way maybe I could do something along these lines:

```go
r := retryablehttp.NewClient()
r.HTTPClient = wrappedClient
r.CheckRetry = msgraph.DefaultCheckRetry

graphClient := msgraph.NewApplicationsClient(tenantID, r.StandardClient())
```

msgraph/client.go
manicminer: A slice of function pointers makes more sense 👍

msgraph/client.go
manicminer: Nice, I like that. It has been asked for, but I was hesitant to overly clutter the `NewClient` function signature

msgraph/client.go
manicminer: This was the intent actually, as the `ConsistencyFailureFunc` is defined per operation / request type, e.g. a consistency error looks different for say `GET /foo/1` to `PUT /foo/1`

msgraph/client.go
manicminer: Although actually, to be used anywhere except corner cases, this would also have to be cascaded to every resource client

msgraph/client.go
manicminer: That would be a nice feature, but the problem is that the default retry behavior sources a `ConsistencyFailureFunc` directly from the input struct for each request, so it can be determined whether the failure was due to replication delay within Azure AD. If you wanted to get the generic retry behavior (e.g. 429 backoff, 5xx retries) you can just use `retryablehttp` as it's providing that functionality here.

PR #81, opened 2021-07-19
Support custom schema extensions for users and groups

manicminer:
This introduces support for schema extension data for Groups and Users

Marshaling of schema extension data is handled automatically by the Group and User structs, enabling use of the existing `Update()` methods on the respective clients.

Unmarshaling is handled by either the provided `msgraph.SchemaExtensionMap` type, or a custom type supplied by the caller. Such a custom type must have an explicit `UnmarshalJSON()` method to satisfy the `SchemaExtensionProperties` interface. Both approaches are tested in the `TestSchemaExtensionsClient()` test.

### DIFF
--- a/internal/cmd/test-cleanup/main.go
+++ b/internal/cmd/test-cleanup/main.go
@@ -56,5 +56,6 @@ func main() {
 	cleanupApplications()
 	cleanupGroups()
 	cleanupUsers()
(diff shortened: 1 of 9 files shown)


MarkDordoy:
This looks great, looking forward to using this in my project

PR #86, opened 2021-08-12
Support for generic directory objects and improved odata field handling

manicminer:
- Adds a top level model struct for the directoryObject API base class
- Embed this struct into other models to inherit the ID and odata fields
- Support operations via the /directoryObjects endpoint
- Improved handling of @odata.bind fields so that the tenant ID isn't needed to build odata URIs for objects

Amongst other benefits, this should enable us to determine the type of an object given its UUID.

### DIFF
--- a/msgraph/directory_objects.go
+++ b/msgraph/directory_objects.go
@@ -37,21 +37,25 @@ func (c *DirectoryObjectsClient) Get(ctx context.Context, id string, query odata
 	if err != nil {
 		return nil, status, fmt.Errorf("DirectoryObjects.BaseClient.Get(): %v", err)
 	}
+
 	defer resp.Body.Close()
 	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
 	}
+
 	var directoryObject DirectoryObject
 	if err := json.Unmarshal(respBody, &directoryObject); err != nil {
 		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
 	}
+
 	return &directoryObject, status, nil
 }
 
 // GetByIds retrieves multiple DirectoryObjects from a list of IDs.
 func (c *DirectoryObjectsClient) GetByIds(ctx context.Context, ids []string, types []odata.ShortType) (*[]DirectoryObject, int, error) {
 	var status int
+
 	body, err := json.Marshal(struct {
 		IDs   []string     `json:"ids"`
 		Types []odata.Type `json:"types"`
@@ -62,6 +66,7 @@ func (c *DirectoryObjectsClient) GetByIds(ctx context.Context, ids []string, typ
 	if err != nil {
 		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
 	}
+
 	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
 		Body:                   body,
 		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
@@ -74,17 +79,20 @@ func (c *DirectoryObjectsClient) GetByIds(ctx context.Context, ids []string, typ
 	if err != nil {
 		return nil, status, fmt.Errorf("DirectoryObjects.BaseClient.Post(): %v", err)
 	}
+
 	defer resp.Body.Close()
 	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
 	}
+
 	var data struct {
 		Objects []DirectoryObject `json:"value"`
 	}
 	if err := json.Unmarshal(respBody, &data); err != nil {
 		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
 	}
+
 	return &data.Objects, status, nil
 }
 
@@ -101,6 +109,7 @@ func (c *DirectoryObjectsClient) Delete(ctx context.Context, id string) (int, er
 	if err != nil {
 		return status, fmt.Errorf("DirectoryObjects.BaseClient.Get(): %v", err)
 	}
+
 	return status, nil
 }
 
@@ -108,6 +117,7 @@ func (c *DirectoryObjectsClient) Delete(ctx context.Context, id string) (int, er
 // id is the object ID of the directory object.
 func (c *DirectoryObjectsClient) GetMemberGroups(ctx context.Context, id string, securityEnabledOnly bool) (*[]DirectoryObject, int, error) {
 	var status int
+
 	body, err := json.Marshal(struct {
 		SecurityEnabledOnly bool `json:"securityEnabledOnly"`
 	}{
@@ -116,6 +126,7 @@ func (c *DirectoryObjectsClient) GetMemberGroups(ctx context.Context, id string,
 	if err != nil {
 		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
 	}
+
 	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
 		Body:                   body,
 		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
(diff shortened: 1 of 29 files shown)


msgraph/directory_objects_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	g.client = msgraph.NewGroupsClient(c.connection.AuthConfig.TenantID)
+	g.client.BaseClient.Authorizer = c.connection.Authorizer
+
+	u := UsersClientTest{
+		connection:   test.NewConnection(auth.MsGraph, auth.TokenVersion2),
+		randomString: rs,
+	}
+	u.client = msgraph.NewUsersClient(c.connection.AuthConfig.TenantID)
+	u.client.BaseClient.Authorizer = c.connection.Authorizer
```
tombuildsstuff, line 41: it might be worth adding some test-wrappers here?

msgraph/directory_objects_test.go
```diff
@@ -0,0 +1,190 @@
...
+		},
+	}
+	group2 := testGroupsClient_Create(t, g, newGroup2)
+
+	testDirectoryObjectsClient_Get(t, c, *user.ID)
+	testDirectoryObjectsClient_Get(t, c, *group1.ID)
+	testDirectoryObjectsClient_GetMemberGroups(t, c, *user.ID, true, []string{*group1.ID, *group2.ID})
+	testDirectoryObjectsClient_GetMemberObjects(t, c, *group1.ID, true, []string{*group2.ID})
+	testDirectoryObjectsClient_GetByIds(t, c, []string{*group1.ID, *group2.ID, *user.ID}, []string{odata.ShortTypeGroup})
+	testDirectoryObjectsClient_Delete(t, c, *group1.ID)
```
tombuildsstuff, line 79: using something like [Testify's `suite` package](https://pkg.go.dev/github.com/stretchr/testify/suite?utm_source=godoc) would make this easier fwiw (since these'd be methods hanging off of that struct with a Setup, as your doing here?)

msgraph/directory_objects_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if len(*directoryObjects) == 0 {
+		t.Fatal("DirectoryObjectsClient.GetMemberGroups(): directoryObjects was empty")
+	}
+
+	expectedCount := len(expected)
+	if len(*directoryObjects) < expectedCount {
+		t.Fatalf("DirectoryObjectsClient.GetMemberGroups(): expected at least %d result. has: %d", expectedCount, len(*directoryObjects))
+	}
+	var expectedFound int
```
tombuildsstuff, line 135: pedantic, but isn't this `actualFound|Count`?

msgraph/models.go
```diff
@@ -46,7 +50,7 @@ type Application struct {
 	DeletedDateTime               *time.Time                `json:"deletedDateTime,omitempty"`
 	DisabledByMicrosoftStatus     interface{}               `json:"disabledByMicrosoftStatus,omitempty"`
 	DisplayName                   *string                   `json:"displayName,omitempty"`
-	GroupMembershipClaims         *[]GroupMembershipClaim   `json:"groupMembershipClaims,omitempty"`
+	GroupMembershipClaims         *[]GroupMembershipClaim   `json:"groupMembershipClaims,omitempty"` // TODO: tag needed?
```
tombuildsstuff, line 53: presumably?

odata/odata.go
```diff
@@ -10,20 +10,72 @@ import (
...
+)
+
+type Id string
+
+func (o *Id) UnmarshalJSON(data []byte) error {
+	var id string
+	if err := json.Unmarshal(data, &id); err != nil {
+		return err
+	}
+	*o = Id(regexp.MustCompile(`/v2/`).ReplaceAllString(id, `/v1.0/`))
```
tombuildsstuff, line 25: do we need to do the same for beta here too?

tombuildsstuff (approved):
a few minor comments but this otherwise LGTM 👍

msgraph/directory_objects.go
```diff
@@ -0,0 +1,191 @@
...
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
+	}
+	var directoryObject DirectoryObject
+	if err := json.Unmarshal(respBody, &directoryObject); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+	return &directoryObject, status, nil
```
katbyte, line 49: i'd personlly like to see a little more whitespace here
```suggestion
	})
	if err != nil {
		return nil, status, fmt.Errorf("DirectoryObjects.BaseClient.Get(): %v", err)
	}
	defer resp.Body.Close()
	
	respBody, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
	}
	
	var directoryObject DirectoryObject
	if err := json.Unmarshal(respBody, &directoryObject); err != nil {
		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
	}
	
	return &directoryObject, status, nil
```

msgraph/directory_objects.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
+	}
+	var data struct {
+		Objects []DirectoryObject `json:"value"`
+	}
+	if err := json.Unmarshal(respBody, &data); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+	return &data.Objects, status, nil
```
katbyte, line 88:
```suggestion
	var status int
	body, err := json.Marshal(struct {
		IDs   []string     `json:"ids"`
		Types []odata.Type `json:"types"`
	}{
		IDs:   ids,
		Types: types,
	})
	if err != nil {
		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
	}
	
	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
		Body:                   body,
		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
		ValidStatusCodes:       []int{http.StatusOK},
		Uri: Uri{
			Entity:      "/directoryObjects/getByIds",
			HasTenantId: true,
		},
	})
	if err != nil {
		return nil, status, fmt.Errorf("DirectoryObjects.BaseClient.Post(): %v", err)
	}
	defer resp.Body.Close()
	
	respBody, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
	}
	
	var data struct {
		Objects []DirectoryObject `json:"value"`
	}
	if err := json.Unmarshal(respBody, &data); err != nil {
		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
	}
	
	return &data.Objects, status, nil
```

msgraph/directory_objects_test.go
katbyte: +1

katbyte (approved):
aside from a couple mindor comments LGTM 👍

msgraph/models.go
```diff
@@ -1118,6 +1110,9 @@ func (u User) MarshalJSON() ([]byte, error) {
 }
 
 func (u *User) UnmarshalJSON(data []byte) error {
+	if u.DirectoryObject == nil {
+		u.DirectoryObject = &DirectoryObject{}
+	}
```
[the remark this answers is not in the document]
manicminer: Good call, that makes much more sense

msgraph/models.go
manicminer: This field is custom marshaled/unmarshaled so this can be a blank tag.

odata/odata.go
manicminer: This is to "fix" the odata IDs returned by the API. They get returned with `v2` as the version slug, which are mostly accepted when sending them back to the API, but I think I had one instance where it wasn't, hence this hack :)

msgraph/directory_objects.go
manicminer: Good call, these grew organically and never got truly tidied up.

msgraph/directory_objects_test.go
manicminer: cool, will take a look at that and improving the test setup sprawl 👍

Issue #92, opened 2021-09-01
EntitlementManagement - AccessPackages


kaovd:
General Issue tracker for everything related to AccessPackage implementation within Identity & Governance Management. Will leave PR in draft state any comments, suggestions or contributions welcome 

https://docs.microsoft.com/en-us/graph/api/resources/entitlementmanagement-root?view=graph-rest-beta

- [x] accessPackage
- [x] accessPackageAssignmentPolicy
- [x] accessPackageCatalog
- [x] accessPackageResourceRequest - Non-Standard Implementation (For adding resources to Catalogs)
- [x]  accessPackageResource - RO Non-Standard

Rest of endpoints do not have Creation/Deletion capabiltity / Directly related to IDAG IaC flows

kaovd:
@manicminer Could you potentially give some of this a  glance on in the draft PR? The first three endpoints are pretty straightforward CRUD, but "accessPackageResourceRequest" endpoint only has List and Create availiable where having some workaround implementation on the SDK would be needed. Specifically take a look at accesspackageresourcerequest.go - In order to delete you need to POST as if it where a Create Method. Also seems to be some lag on the backend potentially if youve just created an AP / Group to waited the Create func for 15s

Return is also a bit weird as a resource creation request != resource, therefore the resource ID (but the ID of the Request) is not returned and while can be figured out via pseduo get method on the corresponding tests file (L92) its a bit weird

Let me know what you think about it,  if we are ok with a few work arounds and impact on downstream implemnetation this is definetly do-able, elsewise will remove accessPackageResourceRequest  for this PR

Tests so far are all good
ok      github.com/manicminer/hamilton/msgraph  27.013s

kaovd:
@petr-stupka Thanks this is extremely useful, understood requests are non standard and have a pseduo delete using admin remove. Have pretty much done exactly what youve said if you look at the PR

 Mustve looked over the execute immediately option - Will add, @manicminer will leave Consistency func in but should shorten time if we are OK to hardcore that? No idea why wouldnt be always on.

Elsewise have an issue with the RoleScope endpoint, from what i currently understand once a request has been done to add something to a Catalog, you call the RoleScope endpoint to add it to an AP - but Theres no deletion capability so i cant implement it? Unless its just not documented properly - would be big help if you know.

petr-stupka:
[...]

PR #93, opened 2021-09-01
Entitlement Management - AccessPackage Types

kaovd:
All API Related to main AccessPackages workflow within IG

#92


manicminer:
Hi @kaovd, thanks for this PR!

The way you've implemented deletes for AccessPackageResourceRequests seems ok. It's arguable whether this should just be omitted and allowed to leak downstream but a thin wrapper like this seems reasonable to make it more appear more restful.

The asynchronous behavior when creating AccessPackageResourceRequests is a bit more challenging. For the initial delay, can we implement retries via a ConsistencyFailureFunc rather than a hard sleep? I haven't used this service before, but is [this](https://github.com/manicminer/hamilton/blob/f731fed6dd465932489980c733d03c180974984f/msgraph/accesspackageresourcerequest_test.go#L92) a reliable way to get the request ID? If so, could this potentially be implemented in the client so that the returned struct can be populated? If this were done, it might be wise to add a boolean argument to the method to toggle that particular step.

kaovd:
The only way I've found to delete a resource is by getting its resource ID once the request has been made, as the Requests endpoint wont return the ID of the Resource it creates which while is a bit annoying - it is how the current documentation describes doing this 
https://docs.microsoft.com/en-us/graph/api/accesspackageresourcerequest-post?view=graph-rest-beta&tabs=http
```
To remove an Azure AD app from a catalog, set the catalogId to be of the ID of the catalog, requestType to be AdminRemove, and the accessPackageResource the resource object to be removed. The resource object can be retrieved using list accessPackageResources.
```

For the flow of that endpoint downstream - on a Delete operation on the client would call the Pseduo-Get on the accessPackageResource to then be able to do the delete operation as seen in the Tests from the line you highlighted. Are you saying this could be rolled into the Create operation on the requests endpoint? While could simplify downstream could be a bit messy if so

Will move to a ConsistencyFailureFunc rather than a hard sleep

manicminer:
> The only way I've found to delete a resource is by getting its resource ID once the request has been made, as the Requests endpoint wont return the ID of the Resource it creates which while is a bit annoying - it is how the current documentation describes doing this
> https://docs.microsoft.com/en-us/graph/api/accesspackageresourcerequest-post?view=graph-rest-beta&tabs=http
> 
> ```
> To remove an Azure AD app from a catalog, set the catalogId to be of the ID of the catalog, requestType to be AdminRemove, and the accessPackageResource the resource object to be removed. The resource object can be retrieved using list accessPackageResources.
> ```

Cool, that's what I understood from your implementation 👍 

 
> For the flow of that endpoint downstream - on a Delete operation on the client would call the Pseduo-Get on the accessPackageResource to then be able to do the delete operation as seen in the Tests from the line you highlighted. Are you saying this could be rolled into the Create operation on the requests endpoint so we store it prehand? While could simplify downstream could be a bit messy if so
> 
> Will move to a ConsistencyFailureFunc rather than a hard sleep

Yeah that was my thinking. If you think this would be flaky or too messy we could not bother with it. We could gate it with a `pollForId` boolean on the Create method if you think consumers would sometimes rather skip this.

kaovd:
Did consider this at first but thought it might've not been acceptable, but if your fine with it will see what i can do with it would simplify the workflow

kaovd:
Think I'm mostly done in terms of raw implementation, pollForId bool works well. fixed bug in Policy Update

> $ go test -race ./msgraph -run="TestAccessPackage*"                                                                                                                                                                                                  [±feature/accessPackages ●]
ok      github.com/manicminer/hamilton/msgraph  15.021s

I haven't introduced a seperate _tests for the accesspackageresource endpoint as it gets covered within the accesspackageresourcerequest tests. Can do so if its needed.

kaovd:
> $ go test -race ./msgraph -run="TestAccessPackage*"                                                                                                                                                [±feature/accessPackages ●●]
ok      github.com/manicminer/hamilton/msgraph  28.060s
  
Taking out of draft although probably a few things to review here but feel like I've got good test coverage - let me know

manicminer:
@kaovd Thanks for your work on this! I unfortunately won't have time to review it today but will make sure it gets in to next week's release.

manicminer (approved):
@kaovd I've made some small changes to tidy up a little bit, and added some nil checks in the clients. Overall this LGTM, the tests seems sensible and are passing! Thanks again! 🚀

PR #96, opened 2021-09-07
Password policies use nullable strings

alexwilcox9:
Changing `PasswordPolicies` to type `StringNullWhenEmpty` so it can be successfully updated

See https://github.com/hashicorp/terraform-provider-azuread/pull/550 for context 

@manicminer what tests if any need to be created/updated for this PR?


manicminer (approved):
@alexwilcox9 Thanks for this fix. We don't have test coverage for every field yet so there's nothing else to update.

PR #97, opened 2021-09-08
Use the right base64 decoder as a JWT is a tuple of URL-safe base64 encoded strings

moriyoshi:
This causes https://github.com/hashicorp/terraform-provider-azuread/issues/558 .

*SIDENOTE:* the implementation doesn't verify the attached signature and this needs to be addressed.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"testing"
 
+	"golang.org/x/oauth2"
+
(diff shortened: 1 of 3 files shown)


manicminer:
Hi @moriyoshi, thanks for spotting and fixing this! I added some tests for `auth.ParseClaims` for some extra safety around this as it has become a critical function.

jackofallops (approved):
LGTM 🚀

PR #99, opened 2021-09-09
Add support for employee org data

michaljirman:
Add support for employee org data to the User model


msgraph/models.go
```diff
@@ -1144,6 +1144,7 @@ type User struct {
 	UsageLocation                   *StringNullWhenEmpty     `json:"usageLocation,omitempty"`
 	UserPrincipalName               *string                  `json:"userPrincipalName,omitempty"`
 	UserType                        *string                  `json:"userType,omitempty"`
+	EmployeeOrgData                 *EmployeeOrgData         `json:"employeeOrgData,omitempty"`
```
manicminer, line 1147: Thanks @michaljirman! Could we move this up to maintain sorting of the field names?

michaljirman: done

manicminer (approved):
Thanks, LGTM!

Will merge later today after 0.28.1 is cut.

PR #100, opened 2021-09-10
Override go-retryablehttp error handling to ensure response object is not discarded

manicminer:
- Firstly, ensure that response is returned in go-retryablehttp.Client.Do() after retries have been exhausted
- Secondly, do not return an error after exhausting retries to prevent net/http from discarding the response
- Response is then returned to msgraph.Client.performRequest() so it can make its own determinations about whether to return an error

### DIFF
--- a/msgraph/client.go
+++ b/msgraph/client.go
@@ -55,6 +55,12 @@ type Uri struct {
 	HasTenantId bool
 }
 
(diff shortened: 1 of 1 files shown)


koikonom (approved):
LGTM

Issue #104, opened 2021-09-16
Identity Governance - Access Reviews


amarbut24:
Running into an issue attempting to interact with `/identityGovernance/accessReviews/definitions`

```
type AccessReviewClient struct {
	BaseClient msgraph.Client
}

func NewAccessReviewClient(tenantId string) *AccessReviewClient {
	return &AccessReviewClient{
		BaseClient: msgraph.NewClient(msgraph.VersionBeta, tenantId),
	}
}

func main() {
	ctx := context.Background()

	clientSecret := os.Getenv("CLIENT_SECRET")
	authConfig := &auth.Config{
		Environment:            environments.USGovernmentL4,
		TenantID:               "<redacted>",
		ClientID:               "<redacted>",
		ClientSecret:           clientSecret,
		EnableClientSecretAuth: true,
	}

	authorizer, err := authConfig.NewAuthorizer(ctx, auth.MsGraph)
	if err != nil {
		log.Fatal(err)
	}

	accessReviewClient := NewAccessReviewClient(authConfig.TenantID)
	accessReviewClient.BaseClient.Authorizer = authorizer
	accessReviewClient.BaseClient.Endpoint = environments.MsGraphUSGovL4Endpoint

	resp, _, _, err := accessReviewClient.BaseClient.Get(ctx, msgraph.GetHttpRequestInput{
		ValidStatusCodes: []int{http.StatusOK},
		DisablePaging:    false,
		Uri: msgraph.Uri{
			Entity: "/identityGovernance/accessReviews/definitions",
			HasTenantId: true,
		},
	})
}
```

The response I get back is 
`"could not read response body: http: read on closed response body"`

Did some digging with the help of @MarkDordoy 
And found what we believe is the source of the issue

When performing this call via https://developer.microsoft.com/
Below is an example response (some details redacted) 

```
{
    "@odata.context": "https://graph.microsoft.us/beta/$metadata#identityGovernance/accessReviews/definitions",
    "@odata.count": 4,
    "value": [
        {
            "id": "<redacted>",
            "displayName": "<redacted>",
            "createdDateTime": "<redacted>",
            "lastModifiedDateTime": "<redacted>",
            "status": "InProgress",
            "descriptionForAdmins": "",
            "descriptionForReviewers": "",
            "customDataProvider": null,
            "customData": null,
            "createdBy": {
                "id": "<redacted>",
                "displayName": "<redacted>",
                "type": null,
                "userPrincipalName": "<redacted>"
            },
[...]

PR #105, opened 2021-09-16
Update OData Count from String to Int Type

amarbut24:
Related to this issue - https://github.com/manicminer/hamilton/issues/104


manicminer:
Thanks for the PR @amarbut24, I went ahead and added a test case to cover the Count field so it's clearer what we're expecting from the API :)

jackofallops (approved):
LGTM 🚀

PR #107, opened 2021-09-21
Support for the appRolesAssignedTo endpoint for service principals

manicminer:
There are two ways to assign app roles, which essentially link and grant consent between some service principal that represents an application with resources to be accessed, and a user or group, or another service principal representing another app, that wishes to access those resources.

The first way, which is supported by the existing client(s), is to assign the app role to the beneficiary principal (user, group or SP).

The second way, which is the recommended method per the docs, is to assign _out_ the app roles from the service principal hosting the resources. This PR adds a client and tests for this approach.

Docs for new endpoint: https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-approleassignedto?view=graph-rest-1.0&tabs=http
Docs for existing endpoint (note the purple box): https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-approleassignments?view=graph-rest-1.0&tabs=http

### DIFF
--- a/msgraph/app_role_assignments.go
+++ b/msgraph/app_role_assignments.go
@@ -3,9 +3,12 @@ package msgraph
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+
+	"github.com/manicminer/hamilton/odata"
 )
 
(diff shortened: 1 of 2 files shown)


jackofallops (approved):
LGTM 🚀

PR #108, opened 2021-09-27
Add ConsistencyFailureFunc for auditLogs/SignIns

MarkDordoy:
I've discovered doing an odata query such as:
```
Filter: fmt.Sprintf("userId eq '%s' and status/ErrorCode eq 0 and (signInEventTypes/any(s:s eq 'nonInteractiveUser') or signInEventTypes/any(s:s eq 'InteractiveUser'))", *user.ID),
```

for the most part works, but occasionally comes back with:

```
Failed to get sign in logs from tenant e4529dd5-4156-46ab-a8b0-********** when searching for user user@testgovtenant.onmicrosoft.us: SignInLogsClient.BaseClient.Get(): unexpected status 400 with OData error: UnknownError: Unsupported Query.
 ```
After chatting with @manicminer  we decided we should look for this and attempt to handle it.

### CODE REVIEW

File: msgraph/sign_in_reports.go [hunk 1]
```diff
@@ -24,9 +24,18 @@ func NewSignInLogsClient(tenantId string) *SignInReportsClient {
 
 // List returns a list of Sign-in Reports, optionally queried using OData.
 func (c *SignInReportsClient) List(ctx context.Context, query odata.Query) (*[]SignInReport, int, error) {
+
```
manicminer, line 27:
```suggestion
```

manicminer (approved):
Thanks @MarkDordoy, this LGTM ☁️

PR #109, opened 2021-09-29
Handle consistency related error that occurs possibly due to app ID collision in another tenant

manicminer:
This occurs when the backing application for the service principal has not fully replicated. We'd usually see an `ErrorServicePrincipalInvalidAppId` here but sometimes this new error occurs within a 403 response.

Related: https://github.com/hashicorp/terraform-provider-azuread/issues/581


jackofallops (approved):
LGTM 🚀

PR #111, opened 2021-10-01
Add support for user's manager

michaljirman:
Add basic support for retrieving and assigning user's manager; was tested.

Reference to the documentation
* https://docs.microsoft.com/en-us/graph/api/user-list-manager?view=graph-rest-1.0&tabs=http
* https://docs.microsoft.com/en-us/graph/api/user-post-manager?view=graph-rest-1.0&tabs=http


manicminer (approved):
Many thanks @michaljirman, this LGTM 👍

PR #112, opened 2021-10-04
Support for OData HTTP headers

manicminer:
- Implement a new way to pass the entire `odata.Query` object as part of request inputs
- Update all existing clients to pass `odata.Query` in full
  - The existing method of passing a `url.Values`map still works, maintains compatibility and can be used for passing non-odata related query parameters
- Support setting OData-related HTTP headers including `OData-Version` / `OData_MaxVersion`, odata-json parameters on the `Accept` header, and the `ConsistencyLevel` header which isn't strictly in the odata 'standard' but heavily related
- Set the `odata.metadata` parameter to `full` when retrieving directory objects to [ensure the `@odata.id` field is returned](https://docs.oasis-open.org/odata/odata-json-format/v4.0/errata03/os/odata-json-format-v4.0-errata03-os-complete.html#_Toc453766619)
  - This dramatically increases the payload volume so we don't default it everywhere

Docs:
1. [OData Protocol](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html)
2. [OData JSON Format](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html)

Related: https://github.com/hashicorp/terraform-provider-azuread/issues/588


msgraph/accesspackageresourcerolescope.go
```diff
@@ -111,16 +111,16 @@ func (c *AccessPackageResourceRoleScopeClient) Create(ctx context.Context, acces
 // Get retrieves a AccessPackageResourceRoleScope.
 func (c *AccessPackageResourceRoleScopeClient) Get(ctx context.Context, accessPackageId string, id string) (*AccessPackageResourceRoleScope, int, error) {
 	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		OData: odata.Query{
+			Expand: odata.Expand{
+				Relationship: "accessPackageResourceRoleScopes",
+				Select:       []string{"accessPackageResourceRole", "accessPackageResourceScope"},
+			},
+			//Filter: fmt.Sprintf("startswith(originId,'%s')", id),
```
tombuildsstuff, line 119: do we need this?

tombuildsstuff (approved):
otherwise LGTM 👍

msgraph/accesspackageresourcerolescope.go
manicminer: Probably not, I think it's a vestige of testing the API

```suggestion
```

PR #120, opened 2021-11-02
Support for autorest.BearerAuthorizer

manicminer:
Also support sourcing an environment from a string value


katbyte (approved):
Thanks @manicminer - LGTM 🔢

PR #123, opened 2021-11-04
Support more APIs, auth refactoring, multi-tenant auth, implement autorest.Authorizer, remove Azure Germany

manicminer:
- Auth package refactoring:
  - Remove the `auth.Api` type and instead use `environments.Api` directly
  - Use the resource URI instead of the friendly name for Azure CLI auth tokens

- Add the `AuxiliaryTokens()` method to the `auth.Authorizer` interface to support obtaining tokens for additional tenants
- Expand support in `auth.AutorestAuthorizerWrapper` to support any `autorest.Authorizer`
  - `autorest.BearerAuthorizer` and `autorest.MultiTenantBearerAuthorizer` are fully supported with access tokens, refresh tokens and expiry
  - Other authorizers can supply access tokens only
- Support auxiliary tenants with client secret and client certificate authorizers

- Implement the `autorest.Authorizer` interface with `auth.CachedAuthorizer` (which wraps all supported Authorizers)
  - This allows authorizers to be used with https://github.com/Azure/go-autorest, with multi-tenant support, with the exception of `auth.MsiAuthorizer`

- Export environment configs for more management plane APIs:
  - Resource Manager
  - Batch Management
  - Data Lake
  - Gallery
  - KeyVault
  - Operational Insights
  - OSS RDBMS
  - Service Bus
  - Service Management (Azure Classic)
  - SQL Database
  - Storage
  - Synapse

- Refactor and tidy up tests for the `msgraph` package

- Say goodbye to Azure Germany 🇩🇪 

⚠️ BREAKING CHANGES:

- The signatures for `auth.NewClientCertificateAuthorizer`, `auth.NewClientSecretAuthorizer` and `auth.NewAzureCliAuthorizer` have changed to accommodate passing additional tenant IDs for multi-tenant authorization.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -94,8 +94,8 @@ func (c *Config) NewAuthorizer(ctx context.Context, api Api) (Authorizer, error)
 }
 
 // NewAutorestAuthorizerWrapper returns an Authorizer that sources tokens from a supplied autorest.BearerAuthorizer
-func NewAutorestAuthorizerWrapper(bearerAuthorizer *autorest.BearerAuthorizer) (Authorizer, error) {
-	return &AutorestAuthorizerWrapper{authorizer: bearerAuthorizer}, nil
+func NewAutorestAuthorizerWrapper(autorestAuthorizer autorest.Authorizer) (Authorizer, error) {
+	return &AutorestAuthorizerWrapper{authorizer: autorestAuthorizer}, nil
 }
 
 // NewAzureCliAuthorizer returns an Authorizer which authenticates using the Azure CLI.
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -98,7 +98,7 @@ func NewAzureCliAuthorizer(ctx context.Context, api environments.Api, tenantId s
 
 // NewMsiAuthorizer returns an authorizer which uses managed service identity to for authentication.
 func NewMsiAuthorizer(ctx context.Context, api environments.Api, msiEndpoint, clientId string) (Authorizer, error) {
-	conf, err := NewMsiConfig(resource(api), msiEndpoint, clientId)
+	conf, err := NewMsiConfig(api.Resource(), msiEndpoint, clientId)
 	if err != nil {
 		return nil, err
 	}
@@ -132,8 +132,8 @@ func NewClientCertificateAuthorizer(ctx context.Context, environment environment
 		ClientID:           clientId,
 		PrivateKey:         x509.MarshalPKCS1PrivateKey(priv),
 		Certificate:        cert.Raw,
-		Resource:           resource(api),
-		Scopes:             scopes(api),
+		Resource:           api.Resource(),
+		Scopes:             []string{api.DefaultScope()},
 		TokenVersion:       tokenVersion,
 	}
 
@@ -148,8 +148,8 @@ func NewClientSecretAuthorizer(ctx context.Context, environment environments.Env
 		AuxiliaryTenantIDs: auxTenantIds,
 		ClientID:           clientId,
 		ClientSecret:       clientSecret,
-		Resource:           resource(api),
-		Scopes:             scopes(api),
+		Resource:           api.Resource(),
+		Scopes:             []string{api.DefaultScope()},
 		TokenVersion:       tokenVersion,
 	}
 
@@ -166,12 +166,4 @@ func TokenEndpoint(endpoint environments.AzureADEndpoint, tenant string, version
 	}
 	e = fmt.Sprintf("%s/token", e)
 	return
-}
-
-func scopes(api environments.Api) (s []string) {
-	return []string{fmt.Sprintf("%s/.default", api.Endpoint)}
-}
-
-func resource(api environments.Api) (r string) {
-	return fmt.Sprintf("%s/", api.Endpoint)
 }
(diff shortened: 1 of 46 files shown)


auth/auth.go
```diff
@@ -170,22 +168,10 @@ func TokenEndpoint(endpoint environments.AzureADEndpoint, tenant string, version
...
-func scopes(env environments.Environment, api Api) (s []string) {
-	switch api {
-	case MsGraph:
-		s = []string{fmt.Sprintf("%s/.default", env.MsGraph.Endpoint)}
-	case AadGraph:
-		s = []string{fmt.Sprintf("%s/.default", env.AadGraph.Endpoint)}
-	}
-	return
+func scopes(api environments.Api) (s []string) {
+	return []string{fmt.Sprintf("%s/.default", api.Endpoint)}
```
tombuildsstuff, line 172: As from https://github.com/hashicorp/go-azure-helpers/pull/83#discussion_r749173708 - we should make a public helper for this - this'll be common enough that you'd want a struct exposing the `Endpoint` and `Resource` for a Resource (and a helper for `DefaultScopeForEndpoint`)?

auth/azcli.go
```diff
@@ -36,29 +86,15 @@ func (a *AzureCliAuthorizer) Token() (*oauth2.Token, error) {
...
-	var resourceType string
-	switch a.conf.Api {
-	case MsGraph:
-		resourceType = "ms-graph"
-	case AadGraph:
-		resourceType = "aad-graph"
-	}
-
-	azArgs := []string{"account", "get-access-token", fmt.Sprintf("--resource-type=%s", resourceType)}
+	azArgs := []string{"account", "get-access-token", fmt.Sprintf("--resource=%s", a.conf.Endpoint)}
```
tombuildsstuff, line 89: since this is a URI (and could be a Resource containing values which should be encoded) - should we quote this value for now?

```suggestion
	azArgs := []string{"account", "get-access-token", fmt.Sprintf("--resource=%q", a.conf.Endpoint)}
```

auth/azcli.go
```diff
@@ -67,45 +103,35 @@ func (a *AzureCliAuthorizer) Token() (*oauth2.Token, error) {
...
+	if cloudShell := os.Getenv("AZUREPS_HOST_ENVIRONMENT"); strings.HasPrefix(cloudShell, "cloud-shell/") {
+		return nil, fmt.Errorf("auxiliary tokens not supported in Cloud Shell")
 	}
 
-	return &AzureCliConfig{Api: api, TenantID: tenantId}, nil
-}
+	tokens := make([]*oauth2.Token, 0)
+	for _, tenantId := range a.conf.AuxiliaryTenantIDs {
+		var token azureCliToken
+		err := jsonUnmarshalAzCmd(&token, "account", "get-access-token", fmt.Sprintf("--resource=%s", a.conf.Endpoint), "--tenant", tenantId)
```
tombuildsstuff, line 123: ditto:

```suggestion
		err := jsonUnmarshalAzCmd(&token, "account", "get-access-token", fmt.Sprintf("--resource=%q", a.conf.Endpoint), "--tenant", tenantId)
```

auth/msi.go
```diff
@@ -82,6 +82,11 @@ func (a *MsiAuthorizer) Token() (*oauth2.Token, error) {
 	return token, nil
 }
 
+// AuxiliaryTokens returns additional tokens for auxiliary tenant IDs, for use in multi-tenant scenarios
+func (a *MsiAuthorizer) AuxiliaryTokens() ([]*oauth2.Token, error) {
+	return nil, fmt.Errorf("not implemented")
```
tombuildsstuff, line 87: we should raise this as `Not Supported` more than not implemented?

internal/test/testing.go
```diff
@@ -54,10 +57,192 @@ func NewConnection(api auth.Api, tokenVersion auth.TokenVersion) *Connection {
...
 	}
 
-	t.Authorizer, err = t.AuthConfig.NewAuthorizer(t.Context, api)
+	return &t
+}
+
+// Authorize configures an Authorizer for the Connection
+func (c *Connection) Authorize(api environments.Api) {
+	var err error
+	c.Authorizer, err = c.AuthConfig.NewAuthorizer(c.Context, api)
```
tombuildsstuff, line 66: this should pass the context onto the methods rather than being on the Connection object here?

tombuildsstuff (approved):
A few minor comments but this otherwise LGTM 👍

auth/auth.go
manicminer: Great idea, have added these as methods on `environments.Api`

auth/azcli.go
manicminer: Reading the docs and a couple of related issues, it looks like exec.Command() performs implicit quoting? i.e. manually quoting seems to produce extra escaped quotes in the arg.

internal/test/testing.go
manicminer: Given this helper is for acceptance testing, I went with moving the context into the top-level `test.Test` struct that gets instantiated at the start of each test, then ensured the method on the `Connection` struct method accepts a ctx argument, and ensured the cancel func is deferred for every test. Although in testing I noticed that the context in `testing.T` seems to always return an empty deadline anyway!

PR #128, opened 2021-11-26
Update endpoints KeyVaultUSGovEndpoint const

MarkDordoy:
Fix type for us gov endpoint


MarkDordoy:
@manicminer FYI

manicminer (approved):
Thanks for the fix @MarkDordoy 👍

PR #129, opened 2021-11-26
Add missing environment API configurations

manicminer:
Also add an `IsAvailable()` convenience method for environments with no support for a particular service/endpoint

### DIFF
--- a/environments/apis.go
+++ b/environments/apis.go
@@ -15,15 +15,15 @@ type Api struct {
 	Endpoint ApiEndpoint
 }
 
-func (a *Api) IsAvailable() bool {
-	return a != nil && *a != ApiUnavailable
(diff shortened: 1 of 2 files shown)


environments/apis.go
```diff
@@ -13,6 +15,10 @@ type Api struct {
 	Endpoint ApiEndpoint
 }
 
+func (a *Api) IsAvailable() bool {
```
tombuildsstuff, line 18: tbh since Api is always a concrete object we could make this:

```suggestion
func (a Api) IsAvailable() bool {
```

and then remove the a != nil check below?

tombuildsstuff (approved):
one comment but this otherwise LGTM 👍

PR #131, opened 2021-12-06
Helper func to load environment from metadata endpoint

manicminer:
Helper func to generate environment configuration from metadata endpoint

- Intended as an equivalent to https://github.com/Azure/go-autorest/blob/master/autorest/azure/metadata_environment.go#L96-L141
- Not entirely useful as you need to supply the Resource Manager endpoint which is then re-used in the output struct, and the metadata endpoint does not currently provide an MS Graph endpoint
- But it should work as a drop in replacement for the above linked helper function in go-autorest

- Also fix an incorrect API ID for Key Vault

### DIFF
--- a/.github/workflows/environments-tests.yml
+++ b/.github/workflows/environments-tests.yml
@@ -1,0 +1,27 @@
+---
+name: Tests - environments
+on:
+  pull_request:
+    types: ["opened", "synchronize"]
+    paths:
+      - "environments/**.go"
+      - ".github/workflows/environments-tests.yml"
+
+jobs:
+  test-environments:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/environments-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+    types: ["opened", "synchronize"]
+    paths:
+      - "environments/**.go"
+      - ".github/workflows/environments-tests.yml"
+
+jobs:
+  test-environments:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
```
jackofallops, line 14: You can remove this, it only affects jobs with a `matrix` configuration.

jackofallops (approved):
One minor comment, but otherwise LGTM 👍

.github/workflows/environments-tests.yml
manicminer: Ah right, thanks! Will fix these in one go 👍

Issue #121, opened 2021-11-03
Support for Federated identity credential


simongottschlag:
Hi @manicminer!

There's a quite new endpoint that would be really awesome to have support for, especially in regard to using it with the Azure AD terraform provider: https://docs.microsoft.com/en-us/graph/api/resources/federatedidentitycredentials-overview?view=graph-rest-beta

The big reason for this is to enable SSO between GitHub Actions and Azure / Azure AD: https://docs.microsoft.com/en-us/azure/developer/github/connect-from-azure?tabs=azure-cli%2Clinux#use-the-azure-login-action-with-openid-connect

manicminer:
Hi @simongottschlag, sorry for the delay in my reply - this looks really interesting! Would you be able to request this in the AzureAD issue tracker? We'll be able to track the potential feature and it will give us a better idea of what this might look like. Thanks!

PR #134, opened 2021-12-21
Support for federated identity credentials for applications

manicminer:
- Support for [Federated Identity Credentials](https://docs.microsoft.com/en-us/graph/api/resources/federatedidentitycredential?view=graph-rest-beta) (beta-only)

Closes: #121


katbyte (approved):
~I think we're missing docs? otherwise~ looks good 🌋 

edit nope ignore that! auto pilot here

PR #139, opened 2022-01-23
Separate autorest module

### LINKED ISSUE #138: Move everything using Microsofts autorest to it's own module
Hi!

As always, love the job you are doing here! :1st_place_medal: 

As per our previous dialouge, it would be quite nice if hamilton could be used without having to import Microsofts autorest library.

I see two possible ways of handling this:

**Choice 1 - new module in hamilton repo**
PROS:

- No need to switch between repositories
- Everything in the same place and easy to find
- Logical to find it here

CONS:

- Go mono-repo support is still quite bad, need to handle a lot of stuff with scripts to keep everything in sync
- Same as above, but we'll need to keep tags like `auth/autorest/v0.39.0` in sync with `v0.39.0`
- Same as above, but we'll need to update the go.mod of the autorest module **after** the main module has been updated/released

**Choice 2 - new repository**

PROS:

- No need for special tooling, will just work
- No need for special tags
- Will work with Dependabot

CONS:

- Will be a little harder to find
- Will have to be updated after hamilton has been updated, but could be handled by dependabot

**Conclusion**

I would personally vote for a separate repository, `hamilton-autorest`, and just place everything that is using autorest there. This repository should be "clean" and easy to use - if autorest is needed it is available in a different place.

simongottschlag:
Fixes #138 

@manicminer this is a small example of how we could move all the autorest dependencies from being imported if using 'pure' hamilton.


go.mod
```diff
@@ -3,8 +3,6 @@ module github.com/manicminer/hamilton
 go 1.16
 
 require (
-	github.com/Azure/go-autorest/autorest v0.11.21
-	github.com/Azure/go-autorest/autorest/adal v0.9.14
```
simongottschlag, line 7: @manicminer very niiiice! (in Borat voice) :1st_place_medal:

simongottschlag:
@manicminer feel free to move this to a separate repository if you want to.

simongottschlag:
Closing as @manicminer is moving it to https://github.com/manicminer/hamilton-autorest/ 🎉⭐️

manicminer:
@simongottschlag Thanks again for your assistance and work on this! 🙏

PR #141, opened 2022-01-23
service/containerregistry: Add ACRHelper

### LINKED ISSUE #140: Add Azure Container Registry helper for docker credentials using hamilton
Hi!

I am currently trying to get `cosign` to work with Azure Container Registry: https://github.com/sigstore/cosign/issues/1350

It is using the following project to get the token for ACR: https://github.com/chrismellard/docker-credential-acr-env

I would like to include a similar helper in hamilton that will be using hamilton instead of autorest.

simongottschlag:
ACRHelper is a credentials.Helper from docker-credential-helpers project
that makes it possible to use it with go-containerregistry.

Fixes #140

### DIFF
--- a/services/containerregistry/acrhelper.go
+++ b/services/containerregistry/acrhelper.go
@@ -1,0 +1,164 @@
+package containerregistry
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+	"github.com/manicminer/hamilton/auth"
+	"github.com/manicminer/hamilton/environments"
+)
+
+type ACRHelper struct{}
+
+var _ credentials.Helper = (*ACRHelper)(nil)
+
+func NewACRHelper() credentials.Helper {
+	return &ACRHelper{}
+}
+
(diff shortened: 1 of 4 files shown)


simongottschlag:
@manicminer this is a draft of how it could be. An alternative is that we make it easy to get the ACR Token using this and put the "wrapper" in another place - to not include any new dependencies for hamilton.

What would you prefer?

services/containerregistry/acrhelper.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if res.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("received non-200 status code: %d", res.StatusCode)
+	}
+
+	var resData struct {
+		RefreshToken string `json:"refresh_token"`
+	}
+
+	err = json.Unmarshal(resBytes, &resData)
```
simongottschlag, line 147: Check error

services/containerregistry/acrhelper.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	resBytes, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
simongottschlag, line 139: Move to before reading body

simongottschlag:
Creating new PR without the helper import.

PR #142, opened 2022-01-24
Support for authenticating via GitHub OIDC

manicminer:
- Add a new authorizer GitHubOIDCAuthorizer which supports [OIDC token exchange](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect) for authenticating to Azure Active Directory

The GHA workflow that tests this runs on a hosted runner and is [already committed](https://github.com/manicminer/hamilton/commit/cdddb2e2e6a7a2af68d8cb9ede7be385eeb44aa4) to main, for testing this PR, and also includes a step that runs the [azure/login](https://github.com/Azure/login) action for validating that the same backend configuration is valid for the GitHubOIDCAuthorizer test. The azure/login action is not needed for this support to work, it's only being run for comparison.

Docs References:
- https://docs.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation
- https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#third-case-access-token-request-with-a-federated-credential
- https://docs.microsoft.com/en-us/azure/developer/github/connect-from-azure?tabs=azure-portal%2Clinux
- https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure

### DIFF
--- a/auth/github.go
+++ b/auth/github.go
@@ -1,0 +1,146 @@
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/oauth2"
+
+	"github.com/manicminer/hamilton/environments"
+)
+
+type GitHubOIDCConfig struct {
+	// Environment is the national cloud environment to use
+	Environment environments.Environment
+
+	// TenantID is the required tenant ID for the primary token
+	TenantID string
+
+	// AuxiliaryTenantIDs is an optional list of tenant IDs for which to obtain additional tokens
+	AuxiliaryTenantIDs []string
+
+	// ClientID is the application's ID.
+	ClientID string
+
+	// IDTokenRequestURL is URL for GitHub's OIDC provider.
+	IDTokenRequestURL string
+
+	// IDTokenRequestToken is the bearer token for the request to the OIDC provider.
+	IDTokenRequestToken string
+
+	// Scopes specifies a list of requested permission scopes (used for v2 tokens)
+	Scopes []string
+
+	// TokenURL is the clientCredentialsToken endpoint, which overrides the default endpoint constructed from a tenant ID
+	TokenURL string
+
+	// Audience optionally specifies the intended audience of the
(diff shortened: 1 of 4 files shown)


auth/github.go
```diff
@@ -0,0 +1,140 @@
...
+	// Scopes specifies a list of requested permission scopes (used for v2 tokens)
+	Scopes []string
+
+	// TokenURL is the clientCredentialsToken endpoint, which overrides the default endpoint constructed from a tenant ID
+	TokenURL string
+
+	// Audience optionally specifies the intended audience of the
+	// request.  If empty, the value of TokenURL is used as the
+	// intended audience.
+	Audience string
```
simongottschlag, line 46: Default should be api://AzureADTokenExchange

auth/github.go
```diff
@@ -0,0 +1,140 @@
...
+	return NewCachedAuthorizer(&GitHubOIDCAuthorizer{ctx, c})
+}
+
+type GitHubOIDCAuthorizer struct {
+	ctx  context.Context
+	conf *GitHubOIDCConfig
+}
+
+func (a *GitHubOIDCAuthorizer) githubAssertion() (*string, error) {
+	req, err := http.NewRequestWithContext(a.ctx, http.MethodPost, a.conf.IDTokenRequestURL, http.NoBody)
```
simongottschlag, line 59: Add audience: https://github.com/actions/toolkit/blob/daf8bb00606d37ee2431d9b1596b88513dcf9c59/packages/core/src/oidc-utils.ts#L72

simongottschlag (commented):
I think the audience required by Azure is missing

manicminer:
Thanks @simongottschlag! I was staring at this for too long and my brain could not do the diff :D

auth/github.go
manicminer: Thanks!

auth/github.go
manicminer: Thanks!

jackofallops (approved):
LGTM 🚀

PR #143, opened 2022-01-28
Groups fixes

manicminer:
- Group model: fix type for `AllowExternalSenders` field
- Don't include the ID in the body when updating a group, as this prevents some Unified group fields from being updated

Test results (note: test runner is currently offline)

Screenshot 2022-01-28 at 00 51 34


katbyte (approved):
LGTM 🌵

PR #144, opened 2022-01-28
services/containerregistry: Add support for Azure Container Registry token exchange

simongottschlag:
This PR introduces `services/containerregistry` and adds support to exchange tokens using an `auth.Authorizer` to an Azure Container Registry token.


simongottschlag:
@manicminer This is a rework of the previous PR, but without the docker import and some tests using a mocked web server.

We had an innovation friday and did some pair programming with @bittrance :^)

services/containerregistry/containerregistry.go
```diff
@@ -0,0 +1,48 @@
...
+	return &ContainerRegistryClient{
+		authorizer,
+		httpClient,
+		serverURL,
+		tenantID,
+	}
+}
+
+// WithHttpClient replaces what http client is used for communication to Azure Container Registry
+func (c *ContainerRegistryClient) WithHttpClient(httpClient *http.Client) {
```
manicminer, line 32: I think it's fine to support user-supplied clients, I'm just curious if you have you found a good use for this other than mocking or debugging?

services/containerregistry/containerregistry.go
```diff
@@ -0,0 +1,48 @@
...
+	if strings.HasPrefix(serverURL, "https://") {
+		scheme = ""
+	}
+
+	serviceURL, err := url.Parse(fmt.Sprintf("%s%s", scheme, serverURL))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse server URL - %w", err)
+	}
+
+	return serviceURL, nil
```
manicminer, line 47: This might be better as?

```suggestion
	serviceURL, err := url.Parse(serverURL)
	if err != nil {
		return nil, fmt.Errorf("failed to parse server URL - %w", err)
	}
	if serviceURL.Scheme == "" {
		serviceURL.Scheme = "https"
	}

	return serviceURL, nil
```

services/containerregistry/refreshtoken.go
```diff
@@ -0,0 +1,112 @@
...
+
+// RefreshTokenClaimsPermissions contains the actions permitted for the refresh token
+type RefreshTokenClaimsPermissions struct {
+	Actions    []string `json:"Actions"`
+	NotActions []string `json:"NotActions"`
+}
+
+// RefreshTokenClaims contians the claims of the refresh token
+// Please observe: These claims are in no way validated, only decoded
+type RefreshTokenClaims struct {
```
manicminer, line 82: Would it be better to use the [`auth.Claims`](https://github.com/manicminer/hamilton/blob/main/auth/claims.go#L12-L27) struct instead of a separate struct here? We can add more fields to it if necessary.

services/containerregistry/refreshtoken.go
```diff
@@ -0,0 +1,112 @@
...
+	Audience       string                        `json:"aud"`
+	Version        string                        `json:"version"`
+	Rid            string                        `json:"rid"`
+	GrantType      string                        `json:"grant_type"`
+	ApplicationID  string                        `json:"appid"`
+	Permissions    RefreshTokenClaimsPermissions `json:"permissions"`
+	Roles          []string                      `json:"roles"`
+}
+
+func decodeRefreshTokenWithoutValidation(token string) (*RefreshTokenClaims, error) {
```
manicminer, line 98: As above, the auth package already has a `ParseClaims` function.

services/containerregistry/accesstoken.go
```diff
@@ -0,0 +1,119 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// ExchangeAccessToken exchanges an Azure Container Registry refresh token for an Azure Container Registry access token
+func (c *ContainerRegistryClient) ExchangeAccessToken(ctx context.Context, refreshToken string, scopes AccessTokenScopes) (string, *AccessTokenClaims, error) {
```
manicminer, line 15: It looks like this can be used as an authorizer in itself, would it be reasonable to implement `auth.Authorizer` with this struct? Along those lines, I think it's good to work with oauth2.Token wherever possible as this helps improve interoperability - can we return an `*oauth2.Token` here instead of a string?

services/containerregistry/refreshtoken.go
```diff
@@ -0,0 +1,112 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// ExchangeRefreshToken exchanges an Azure AD token for an Azure Container Registry refresh token
+func (c *ContainerRegistryClient) ExchangeRefreshToken(ctx context.Context) (string, *RefreshTokenClaims, error) {
```
manicminer, line 15: As per above comment on implementing auth.Authorizer, can we return `*oauth2.Token, error` here? Also, this might be better as `ObtainRefreshToken`?

manicminer (commented):
Thanks for your work on this @bittrance and @simongottschlag! This is off to a great start and I think fits well in the project - I would like to be able to support different services that integrate with AAD.

I've made some comments inline below on considering implementing the `auth.Authorizer` interface here. In line with that, I think this would be better placed as a package under `auth`, i.e. `auth/containerregistry` - WDYT?

Really appreciate the effort put into testing this, on balance I do think that we need to include integration/acceptance tests that work with a real instance of the service. I'll look into getting this set up on the testing tenant to make CI easier.

services/containerregistry/containerregistry.go
simongottschlag: Ah, good! 👍 I'm using it here for mocking. Two use cases are tracing / logging and additional authentication (like client certificate). Don't know how useful those are in this case, but seems like something one should do to be a bit more pluggable for custom use cases.

services/containerregistry/containerregistry.go
simongottschlag: For sure! Will fix.

services/containerregistry/refreshtoken.go
simongottschlag: I completely agree. Will look into what claims it contains and see if it's possible to reuse.

In this case, the only real value is the exp (as of now).

services/containerregistry/refreshtoken.go
simongottschlag: I agree! Will look at it 👍

services/containerregistry/accesstoken.go
simongottschlag: That would be the best way! But important that it doesn't get confused with the Azure AD authorizers, since it won't do much good over there. Will take a look and see if I can come up with something good! 👍

services/containerregistry/refreshtoken.go
simongottschlag: Maybe, I'm not sure yet. It's not implemented as a standard OAuth2 endpoint (at least to my understanding - may be wrong).

Naming is hard and I am bad at it. I welcome any and all name suggestions 😁

simongottschlag:
Hi @manicminer and thank you for taking the time to look at it!

Didn't see your review until after I had turned off the computer.

I agree with you and will try to implement it as you suggest. Will most likely need a few pointers from you to get it right. 😊

My thought is that I'm going to implement some smaller mocks around the base logic and then also implement some end to end tests. I also started implementing the CatalogClient and how these two interact is something I think you should take a look at and decide. I'm currently just adding logic and tests to verify the functionality but should, before we merge, make sure it's following your style of coding and testing.

Hopefully that will be easy when all the logic is implemented and we can move it around.

One thing I'm missing is functional options which would fit quite well with these kinds of clients. As an example, WithHttpClient could be an option instead of a method.

I'll try to update the code based on your initial comments in the coming days. 👍

services/containerregistry/containerregistry.go
simongottschlag: @manicminer doing it your way will actually cause a `serverURL` with the string `foo.bar` to have an empty `Host` and a Path with `foo.bar`.

services/containerregistry/refreshtoken.go
simongottschlag: As by the previous comment, if we're not using the same type for the Claims, we won't be able to use the same parse function until go 1.18 and generics.

services/containerregistry/accesstoken.go
simongottschlag: Regarding authorizer: I've been looking at `auth.Authorizer` as a way to obtain Azure AD tokens, not tokens for Azure Container Registry. In this case, it could be confusing to implement the interface since it's also used by itself to obtain AAD tokens.

Regarding oauth2.Token: I'm not sure Azure Container Registry is implementing the flow in a way that's supported by oauth2.Token. But to be honest, I haven't used it that much myself yet. I would be able to populate it with content, but not actually sure if it would serve any greater purpose since we wouldn't be able to use it for refreshing the token (as far as I understand).

Any feedback on the above two points is appriciated. I'm not sure about them and it may be worth you looking at it a bit deeper to see if it's applicable or not - I may be missing something here.

services/containerregistry/refreshtoken.go
simongottschlag: As a continuation on the previous comment, I'll wait with this until you've taken a look at if you think `*oauth2.Token` is applicable here. When it comes to naming, `Exchange` is what they used in the Azure Go SDK for this. Naming is hard..

services/containerregistry/containerregistry.go
manicminer: Ah I see, that makes sense. Interestingly printing such a `url.URL` still yields a valid URL string :/
https://go.dev/play/p/gkaYgWvOvkm

Happy with your approach!

services/containerregistry/refreshtoken.go
manicminer: Sounds good - if the claims are sufficiently semantically different between access and refresh tokens, then I agree it makes to split them. But I think we should implement this in the `auth` package. Perhaps if there is sufficient overlap we can embed the common claims.

services/containerregistry/refreshtoken.go
manicminer: Sounds good, as above I think parsing refresh tokens should be a common feature of the `auth` package to avoid reimplementing it for subsequent services.

services/containerregistry/accesstoken.go
manicminer: In my mind, `oauth2.Token` is just a familiar container type for HTTP authorization tokens, with some helper methods, which is accepted by many packages. 

The `auth.Authorizer` interface is likewise intended to be somewhat generic, guaranteeing that types will be able to supply a primary token for the Authorization header, and any supplementary tokens where applicable for the `X-Ms-Authorization-Auxiliary` header. I think it's fine to accept `Authorizer`s for use in other clients like ACR - how the token is obtained and what it's valid for is largely orthogonal to that :)

If a client wishes to carry out some validation on tokens provided by an Authorizer to determine suitability, parsing the claims should yield sufficient data for that - e.g. comparing the `aud` claim.

services/containerregistry/refreshtoken.go
simongottschlag: @manicminer I think the RefreshTokenClaims here are to specific to Container Registry to be used else where, as the `permissions` won't be implemented (most likely) in any other API. I have no problem moving it to `auth`, but I do think it may confuse others that it's in some way related to an Azure AD refresh token wich it isn't.

simongottschlag:
@manicminer I've now also added support for `auth.Authorizer`. I've kept it separate from the `auth` package for now, but if you want to move it I'll do it. :)

PR #145, opened 2022-01-30
Run tests on cloud runners

manicminer:
- Support for selecting GitHub OIDC authentication when using the `auth.NewAuthorizer()` helper function
- Bump supported Go version to 1.17.6

---

Run all tests on cloud hosted GHA runners.

msgraph tests now use GitHub OIDC authentication to acquire a token for creating/deleting test resources.

Uses different credentials for testing client credentials auth flow and Azure CLI auth - as these are now stored in GH secrets they can potentially be leaked but they have no meaningful roles or scopes assigned so are powerless to do anything.

### DIFF
--- a/.github/workflows/auth-tests.yml
+++ b/.github/workflows/auth-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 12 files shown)


jackofallops (approved):
LGTM 👍

Issue #118, opened 2021-10-25
Add support for `claimsMappingPolicy` 


tiwood:
To be able to manage application claims mappings with Terraform we need to be able to use the `claimsMappingPolicy` API in `hamilton`.

## References

* https://github.com/hashicorp/terraform-provider-azuread/issues/644
* https://docs.microsoft.com/en-us/graph/api/resources/claimsmappingpolicy?view=graph-rest-1.0
* https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-claimsmappingpolicies?view=graph-rest-1.0&tabs=http

computeracer:
I was taking a look at implementing support for this, but how do you suggest handling the `definition` field of a potential claims mapping policy struct?

Looking at the properties of a definition as defined in https://docs.microsoft.com/en-us/graph/api/resources/claimsmappingpolicy?view=graph-rest-1.0#properties-of-a-claims-mapping-policy-definition there are complex JSON object types. Are there any examples in the current code base for implemented resources that are similar?

It seems like we would need to implement a custom unmarshal and marshal json functions as well as possibly declare custom types for each of the definition properties. A naive implementation could just treat definition as an array of strings, and we ensure json encoded content, but looking for ideas :smile: 

I am open to any input on implementation. Thank you!

manicminer:
@computeracer In this case the policy is just sent as a string, so for a first pass I'd suggest typing this as a string. The client will have to handle building the policy, using whatever data types make sense for the use case. Similarly for unmarshaling, just treat it as a string and return the field as such. No need to validate it either, just let the API do the work.

PR #147, opened 2022-02-01
:sparkles: Add ClaimsMappingPolicy

computeracer:
This adds initial support for the ClaimsMappingPolicy resource.
Support for this resource type is prerequisite for adding support
to the azure ad terraform provider so it can support Azure AD
Claims Mapping polices.

Fixes: manicminer/hamilton#118

Graph API Reference:
https://docs.microsoft.com/en-us/graph/api/resources/claimsmappingpolicy?view=graph-rest-1.0

### DIFF
--- a/msgraph/claims_mapping_policy.go
+++ b/msgraph/claims_mapping_policy.go
@@ -1,0 +1,169 @@
+package msgraph
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/manicminer/hamilton/odata"
+)
+
+type ClaimsMappingPolicyClient struct {
+	BaseClient Client
+}
+
+// NewClaimsMappingPolicyClient returns a new ClaimsMappingPolicyClient
+func NewClaimsMappingPolicyClient(tenantId string) *ClaimsMappingPolicyClient {
+	return &ClaimsMappingPolicyClient{
+		BaseClient: NewClient(VersionBeta, tenantId),
+	}
+}
+
+// Create creates a new ClaimsMappingPolicy.
+func (c *ClaimsMappingPolicyClient) Create(ctx context.Context, policy ClaimsMappingPolicy) (*ClaimsMappingPolicy, int, error) {
+	var status int
+
+	body, err := json.Marshal(policy)
+	if err != nil {
+		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
+	}
+
+	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
+		Body:             body,
+		ValidStatusCodes: []int{http.StatusCreated},
+		Uri: Uri{
+			Entity:      "/policies/claimsMappingPolicies",
+			HasTenantId: false,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("ClaimsMappingPolicyClient.BaseClient.Post(): %v", err)
+	}
+
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
+	}
+
+	var newPolicy ClaimsMappingPolicy
+	if err := json.Unmarshal(respBody, &newPolicy); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+
+	return &newPolicy, status, nil
+}
+
+// List returns a list of ClaimsMappingPolicy, optionally queried using OData.
+func (c *ClaimsMappingPolicyClient) List(ctx context.Context, query odata.Query) (*[]ClaimsMappingPolicy, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		DisablePaging:    query.Top > 0,
+		OData:            query,
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      "/policies/claimsMappingPolicies",
+			HasTenantId: false,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("ClaimsMappingPolicyClient.BaseClient.Get(): %v", err)
+	}
+
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
+	}
+
+	var data struct {
+		ClaimsMappingPolicies []ClaimsMappingPolicy `json:"value"`
+	}
+	if err := json.Unmarshal(respBody, &data); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+
+	return &data.ClaimsMappingPolicies, status, nil
+}
+
+// Get retrieves a ClaimsMappingPolicy.
+func (c *ClaimsMappingPolicyClient) Get(ctx context.Context, id string, query odata.Query) (*ClaimsMappingPolicy, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
+		OData:                  query,
+		ValidStatusCodes:       []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/policies/claimsMappingPolicies/%s", id),
+			HasTenantId: false,
+		},
+	})
+	if err != nil {
+		return nil, status, fmt.Errorf("ClaimsMappingPolicyClient.BaseClient.Get(): %v", err)
+	}
+
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, status, fmt.Errorf("io.ReadAll(): %v", err)
+	}
+
+	var claimsMappingPolicies ClaimsMappingPolicy
+	if err := json.Unmarshal(respBody, &claimsMappingPolicies); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+
+	return &claimsMappingPolicies, status, nil
+}
+
+// Update amends an existing ClaimsMappingPolicy.
+func (c *ClaimsMappingPolicyClient) Update(ctx context.Context, claimsMappingPolicy ClaimsMappingPolicy) (int, error) {
+	var status int
+
+	if claimsMappingPolicy.ID == nil {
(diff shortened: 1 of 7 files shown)


computeracer:
@manicminer I left this as draft until we implement the claimsMappingPolicy assignment to service principal.

webframp:
@manicminer We'd appreciate any feedback you have on this now that the policy assignment support is added. We're happy to clean this up based on feedback at your convenience.

We plan to use this to implement support in the azuread provider next to support hashicorp/terraform-provider-azuread#644

webframp:
@manicminer I checked out the error but it looks unrelated:

```
    testing.go:149: could not acquire access token: githubAssertion: cannot request token: Get "?audience=api%3A%2F%2FAzureADTokenExchange": unsupported protocol scheme ""
```

I haven't seen this in our testing and I'm able to run all tests locally with my dev subscription. I'm happy to update anything you need though to fix this, just let me know what contribution process you want to follow with this repo.

webframp:
Awesome, appreciate your work maintaining this library. 

I'll push up a rebase soon and am also available on the slack as needed.

webframp:
Rebased on main but testing still seems to fail here, let me know how else we can help as needed. 

Here's the local results I get after our changes but it's possible we missed something:
```
❯ go test -timeout 30s -run ^TestClaimsMappingPolicyClient$ github.com/manicminer/hamilton/msgraph -v
=== RUN   TestClaimsMappingPolicyClient
--- PASS: TestClaimsMappingPolicyClient (3.48s)
PASS
ok      github.com/manicminer/hamilton/msgraph  3.492s
```

Thanks again whenever you get the chance to look at this.

msgraph/serviceprincipals.go
```diff
@@ -348,6 +348,136 @@ func (c *ServicePrincipalsClient) RemoveOwners(ctx context.Context, servicePrinc
...
+		// don't fail if an owner already exists
+		checkPolicyAlreadyExists := func(resp *http.Response, o *odata.OData) bool {
+			if resp != nil && resp.StatusCode == http.StatusBadRequest && o != nil && o.Error != nil {
+				return o.Error.Match(odata.ErrorAddedObjectReferencesAlreadyExist)
+			}
+			return false
+		}
+
+		path := odata.Id(fmt.Sprintf("https://graph.microsoft.com/v1.0/policies/claimsMappingPolicies/%s", *policy.ID))
+		body, err := json.Marshal(ClaimsMappingPolicy{ODataId: &path})
```
manicminer, line 372: Whilst the handling of $ref endpoints is a bit clunky at the moment, the construction of the OData URI can be left to the odata package and this can be reduced to:

```suggestion
		body, err := json.Marshal(DirectoryObject{ODataId: policy.ODataId})
```

This does require that the `ODataId` field is set on the model, however this can be accomplished by embedding the `DirectoryObject` struct and requesting the full metadata in the `ClaimsMappingPolicyClient` (see next comments).

msgraph/claims_mapping_policy.go
```diff
@@ -0,0 +1,169 @@
...
+	var status int
+
+	body, err := json.Marshal(policy)
+	if err != nil {
+		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
+	}
+
+	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
+		Body:             body,
+		ValidStatusCodes: []int{http.StatusCreated},
```
manicminer, line 35: You can request the full metadata in a POST request, and the response, whilst larger, will contain the all important `@odata.id`. For example:

```suggestion
		OData:            odata.Query{Metadata: odata.MetadataFull},
		ValidStatusCodes: []int{http.StatusCreated},
```

Alas the `@odata.id` returned by the API can arrive in a couple different formats, one of which paradoxically isn't accepted by the API - so the `odata.Id` type has its own workaround for this in its [`MarshalJSON()` method](https://github.com/manicminer/hamilton/blob/main/odata/odata.go#L17-L40). Whilst testing this I noticed the regex doesn't work for nested object URIs, so I've opened a PR #152 to fix that, which you'll need.

msgraph/claims_mapping_policy.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := json.Unmarshal(respBody, &newPolicy); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+
+	return &newPolicy, status, nil
+}
+
+// List returns a list of ClaimsMappingPolicy, optionally queried using OData.
+func (c *ClaimsMappingPolicyClient) List(ctx context.Context, query odata.Query) (*[]ClaimsMappingPolicy, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
```
manicminer, line 61: For completeness, you can also request full metadata here

```suggestion
	query.Metadata = odata.MetadataFull

	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
```

msgraph/claims_mapping_policy.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := json.Unmarshal(respBody, &data); err != nil {
+		return nil, status, fmt.Errorf("json.Unmarshal(): %v", err)
+	}
+
+	return &data.ClaimsMappingPolicies, status, nil
+}
+
+// Get retrieves a ClaimsMappingPolicy.
+func (c *ClaimsMappingPolicyClient) Get(ctx context.Context, id string, query odata.Query) (*ClaimsMappingPolicy, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
```
manicminer, line 92: ...and here

```suggestion
	query.Metadata = odata.MetadataFull

	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
```

(Incidentally we might make this the default setting in future)

msgraph/models.go
```diff
@@ -540,6 +540,22 @@ type BaseNamedLocation struct {
 	ModifiedDateTime *time.Time  `json:"modifiedDateTime,omitempty"`
 }
 
+type ClaimsMappingPolicy struct {
+	ODataId               *odata.Id `json:"@odata.id,omitempty"`
```
manicminer, line 544: Lastly, instead of defining this field here, it's better to embed the `DirectoryObject` struct unless it doesn't make sense to. e.g.

```suggestion
	DirectoryObject

```

msgraph/models.go
```diff
@@ -540,6 +540,22 @@ type BaseNamedLocation struct {
...
+	IsOrganizationDefault *bool     `json:"isOrganizationDefault,omitempty"`
+}
+
+func (o *ClaimsMappingPolicy) Uri(endpoint environments.ApiEndpoint, apiversion ApiVersion) string {
+	if o.ID == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s/%s/policies/claimsMappingPolicies/%s", endpoint, apiversion, *o.ID)
+}
+
```
manicminer, line 558: I don't think we need this method tbh

```suggestion
```

msgraph/models.go
```diff
@@ -1250,7 +1266,8 @@ type ScopedRoleMembership struct {
 // ServicePrincipal describes a Service Principal object.
 type ServicePrincipal struct {
 	DirectoryObject
-	Owners *Owners `json:"owners@odata.bind,omitempty"`
+	Owners                *Owners `json:"owners@odata.bind,omitempty"`
+	ClaimsMappingPolicies *[]ClaimsMappingPolicy `json:"claimsmappingpolicies@odata.bind,omitempty"`
```
manicminer, line 1270: Linter picked this one up, the field name, type and tag should be vertically aligned :)

```suggestion
	Owners                *Owners                `json:"owners@odata.bind,omitempty"
	ClaimsMappingPolicies *[]ClaimsMappingPolicy `json:"claimsmappingpolicies@odata.bind,omitempty"`
```

manicminer (changes requested):
Hey @computeracer, @webframp, sorry for the delay in reviewing. This is looking great! I made a few comments inline regarding the handling of OData IDs / URIs - I think this can be unified a little with the other clients which will make this easier to refactor/improve in the future. If you can take a look at these, this should be good to merge. Thanks!

msgraph/claims_mapping_policy.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+type ClaimsMappingPolicyClient struct {
+	BaseClient Client
+}
+
+// NewClaimsMappingPolicyClient returns a new ClaimsMappingPolicyClient
+func NewClaimsMappingPolicyClient(tenantId string) *ClaimsMappingPolicyClient {
+	return &ClaimsMappingPolicyClient{
+		BaseClient: NewClient(VersionBeta, tenantId),
```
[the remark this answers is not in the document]
computeracer: Moved to 1.0. Thank you.

msgraph/claims_mapping_policy.go
computeracer: Fixed. Thank you.

msgraph/models.go
computeracer: Its not. We removed.

msgraph/models.go
computeracer: Thank you. Fixed the formatting.

msgraph/models.go
computeracer: Got it. Thank you.

msgraph/claims_mapping_policy.go
computeracer: We rebased on the bug fix branch to pull that change in. Our tests locally are passing. Thank you.

msgraph/models.go
```diff
@@ -540,6 +540,15 @@ type BaseNamedLocation struct {
 	ModifiedDateTime *time.Time  `json:"modifiedDateTime,omitempty"`
 }
 
+type ClaimsMappingPolicy struct {
+	DirectoryObject
+	Definition            *[]string `json:"definition,omitempty"`
+	Description           *string   `json:"description,omitempty"`
+	DisplayName           *string   `json:"displayName,omitempty"`
+	ID                    *string   `json:"id,omitempty"`
```
manicminer, line 548: Can we remove the `ID` field from the `ClaimsMappingPolicy` type as this will enable the `ID` field to be embedded from `DirectoryObject`

manicminer (commented):
Hey @computeracer, sorry for the delay in following up here as I've been away for a week. This is looking great! One minor thing to look at is the `ID` field of the `ClaimsMappingPolicy` model struct (as commented inline). I was unfortunately unable to push to your branch or I'd have taken care of that, but if you can look at that then this will be good to merge. Thanks again for the contribution!

msgraph/models.go
computeracer: Thank you for the feedback. This has been adjusted @manicminer.

manicminer (approved):
Thanks for updating!

PR #157, opened 2022-04-14
Fix App ID URI for Vault in USGovernment cloud

manicminer:
Reported in https://github.com/hashicorp/terraform-provider-azurerm/issues/16291


jackofallops (approved):
LGTM 👍

PR #158, opened 2022-04-14
Service Principal: Support token signing certificates

manicminer:
Replaces #151 originally opened by @dhohengassner as I'm unable to push to their fork.

---

✨ add servicePrincipal methods AddTokenSigningCertificate and SetPreferredTokenSigningKeyThumbprint

This commit adds support to create the certificiate for Azure AD signed certs and set
the preferred token thumbprint on the service principal.

This will allow to follow the steps described in
https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api#create-a-signing-certificate
using hamilton SDK.

Currently Microsoft does not support a method to remove the created certificate key from the service principal.
https://docs.microsoft.com/en-us/graph/api/serviceprincipal-addtokensigningcertificate

Also manual removal via `removePassword` and `removekey` API calls are not supported and fail with an internal server error.

This SDK extension is the base to extend the `terraform-provider-azuread`. 

Issue: https://github.com/hashicorp/terraform-provider-azuread/issues/732

### DIFF
--- a/msgraph/models.go
+++ b/msgraph/models.go
@@ -1052,6 +1052,7 @@ type KeyCredential struct {
 	EndDateTime         *time.Time         `json:"endDateTime,omitempty"`
 	KeyId               *string            `json:"keyId,omitempty"`
 	StartDateTime       *time.Time         `json:"startDateTime,omitempty"`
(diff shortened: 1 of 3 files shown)


jackofallops (approved):
LGTM 👍

PR #166, opened 2022-05-31
feat: support oidc federated authentication

ekristen:
## Overview

This implements federated authentication via OIDC. This works for Kubernetes Federation along with Generic OIDC issuer.

The client credentials as pointed out by another individual who's name I cannot find (sorry!) 

This basically allows for any OIDC issuer that is configured with a service principal to work. Simple pass the JWT token from the OIDC issuer to the auth config instead and enable client federated auth. 

## Testing

I did not provide tests as I have no way of setting up federated authentication for those test to pass properly for you but I can walk you through the process if interested, there are a couple ways to do this.

## Example Code

```golang
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/manicminer/hamilton/auth"
	"github.com/manicminer/hamilton/environments"
	"github.com/manicminer/hamilton/msgraph"
	"github.com/manicminer/hamilton/odata"
)

var (
	tenantId = os.Getenv("AZURE_TENANT_ID")
	clientId = os.Getenv("AZURE_CLIENT_ID")
	jwtToken = os.Getenv("AZURE_FEDERATED_TOKEN")
)

func main() {
	ctx := context.Background()

	environment := environments.Global

	authConfig := &auth.Config{
		Environment:               environment,
		TenantID:                  tenantId,
		ClientID:                  clientId,
		FederatedAssertion:        jwtToken,
		EnableClientFederatedAuth: true,
	}

	authorizer, err := authConfig.NewAuthorizer(ctx, environment.MsGraph)
	if err != nil {
		log.Fatal(err)
	}

	client := msgraph.NewUsersClient(tenantId)
	client.BaseClient.Authorizer = authorizer

	users, _, err := client.List(ctx, odata.Query{})
	if err != nil {
		log.Println(err)
		return
	}
	if users == nil {
		log.Println("bad API response, nil result received")
		return
	}
	for _, user := range *users {
		fmt.Printf("%s: %s <%s>\n", *user.ID, *user.DisplayName, *user.UserPrincipalName)
	}
}
```

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -61,6 +61,16 @@ func (c *Config) NewAuthorizer(ctx context.Context, api environments.Api) (Autho
 		}
 	}
 
+	if c.EnableClientFederatedAuth {
+		a, err := NewClientFederatedAuthorizer(ctx, c.Environment, api, c.Version, c.TenantID, c.AuxiliaryTenantIDs, c.ClientID, c.FederatedAssertion)
+		if err != nil {
+			return nil, fmt.Errorf("could not configure ClientCertificate Authorizer: %s", err)
+		}
+		if a != nil {
+			return a, nil
+		}
+	}
+
 	if c.EnableGitHubOIDCAuth {
 		a, err := NewGitHubOIDCAuthorizer(context.Background(), c.Environment, api, c.TenantID, c.AuxiliaryTenantIDs, c.ClientID, c.IDTokenRequestURL, c.IDTokenRequestToken)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ekristen:
I don't believe the azure-cli-tests failed because of my changes, I suspect since its a fork, it can't use the GitHub Action secrets properly. Let me know if there's anything else I can do.

ekristen:
@manicminer any chance of getting eyes on this? Thanks!

manicminer (approved):
Hi @ekristen, thanks for proposing this and apologies for the delay in reviewing as I've been periodically out of office.

This looks good, we'll omit the tests for now as it's nontrivial to supply a valid ID token programmatically. We'll lean on the GHA based test for now since it's doing the same thing.

ekristen:
Nice! Thanks @manicminer -- going to look at Terraform inclusion next.

PR #167, opened 2022-06-26
Add Synchronization client

JoostvDoorn:
These changes are required for the azuread terraform provider changes in PR: https://github.com/hashicorp/terraform-provider-azuread/pull/830

This implements the calls documented here: https://docs.microsoft.com/en-us/graph/api/resources/synchronization-overview?view=graph-rest-beta

### DIFF
--- a/msgraph/synchronization.go
+++ b/msgraph/synchronization.go
@@ -3,12 +3,10 @@ package msgraph
 import (
 	"context"
 	"encoding/json"
-	"errors"
+	// "errors"
 	"fmt"
 	"io"
 	"net/http"
-
-	"github.com/manicminer/hamilton/odata"
 )
 
 // SynchronizationJobClient performs operations on SynchronizationJobs.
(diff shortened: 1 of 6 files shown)


msgraph/synchronization.go
```diff
@@ -0,0 +1,318 @@
...
+
+// SynchronizationJobClient performs operations on SynchronizationJobs.
+type SynchronizationJobClient struct {
+	BaseClient Client
+}
+
+// NewSynchronizationJobClient returns a new SynchronizationJobClient
+func NewSynchronizationJobClient(tenantId string) *SynchronizationJobClient {
+	return &SynchronizationJobClient{
+		BaseClient: NewClient(VersionBeta, tenantId),
```
manicminer, line 20: It looks like synchronization jobs are supported on the 1.0 API - if these work sufficiently can we change this default to target v1.0 instead?

msgraph/synchronization.go
```diff
@@ -0,0 +1,318 @@
...
+
+	body, err := json.Marshal(synchronizationJob)
+	if err != nil {
+		return nil, status, fmt.Errorf("json.Marshal(): %v", err)
+	}
+	resp, status, _, err := c.BaseClient.Post(ctx, PostHttpRequestInput{
+		Body:             body,
+		ValidStatusCodes: []int{http.StatusCreated},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/servicePrincipals/%s/synchronization/jobs/", servicePrincipalId),
```
manicminer, line 148:
```suggestion
			Entity:      fmt.Sprintf("/servicePrincipals/%s/synchronization/jobs", servicePrincipalId),
```

manicminer (changes requested):
Thanks for this contribution @JoostvDoorn! This is mostly looking great, just a couple of very minor things I noticed, and a question about the API version. If you can take a look at these, this should be good to merge, thanks!

msgraph/synchronization.go
JoostvDoorn: Unfortunately it doesn't work for version 1.0, I have tested some of the calls in the graph explorer for version 1.0 and they all gave me errors. Currently the documentation also is only available for the beta API.

msgraph/synchronization.go
JoostvDoorn: Fixed in b574351989a455ba92195ee57c8fc6a5008ecbba

msgraph/synchronization.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return &synchronizationJob, status, nil
+}
+
+// GetSecrets retrieves a SynchronizationSecret
+func (c *SynchronizationJobClient) GetSecrets(ctx context.Context, servicePrincipalId string) (*SynchronizationSecret, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ValidStatusCodes: []int{http.StatusOK},
+		Uri: Uri{
+			Entity:      fmt.Sprintf("/servicePrincipals/%s/synchronization/secrets/", servicePrincipalId),
```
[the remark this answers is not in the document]
JoostvDoorn: Fixed in b574351989a455ba92195ee57c8fc6a5008ecbba

JoostvDoorn:
@manicminer I fixed your comments, and I fixed some consistency errors, which also allowed me to remove the sleeps from the tests. If you see any more issues let me know.

JoostvDoorn:
@manicminer I think CI is ready to be re-run. I have tested multiple times and I haven't run into the `UnableToObtainLockOnResource: Unable to obtain lock on the resource` that the test failed on in the [previous version](https://github.com/manicminer/hamilton/runs/7090667028?check_suite_focus=true#step:4:75).

manicminer (approved):
Hi @JoostvDoorn, thanks for circling back. My apologies for the delay in re-reviewing. This is looking good, I'm going to go ahead and merge now with a view to releasing a little later today. Thanks again :)

PR #173, opened 2022-08-30
Add missing checks for enabling OIDC auth

manicminer:
This checks that required parameters are set to avoid attempting to enable the wrong OIDC method and then erroring prematurely


katbyte (approved):
LGTM 🏥

PR #175, opened 2022-09-05
add app IDs for Front Door

t3mi:
Add known application IDs for Azure Front Door
- Legacy https://docs.microsoft.com/en-us/azure/frontdoor/front-door-custom-domain-https#register-azure-front-door
- Standard/Premium https://docs.microsoft.com/en-us/azure/frontdoor/standard-premium/how-to-configure-https-custom-domain#register-azure-front-door


environments/published.go
```diff
@@ -74,6 +74,8 @@ var PublishedApis = map[string]ApiAppId{
 	"MileIqDashboard":                   "f7069a8d-9edc-4300-b365-ae53c9627fc4",
 	"MileIqRestService":                 "b692184e-b47f-4706-b352-84b288d2d9ee",
 	"MixedReality":                      "c7ddd9b4-5172-4e28-bd29-1e0792947d18",
+	"Microsoft.Azure.Frontdoor"          "ad0e1c7e-6d38-4ba4-9efd-0bc77ba9f037",
+	"Microsoft.AzureFrontDoor-Cdn"       "205478c0-bd83-4e1b-a9d6-db63a3e1e1c8",
```
manicminer, line 78: Could these be pascal-case without punctuation like the other entries?

manicminer (changes requested):
@t3mi Thanks for these additions. I'm happy to add these but have a small request about the formatting (below) if you can take a look? Thanks!

environments/published.go
t3mi: Changed. But don't you think that it's easier for users to use official application names instead of looking how its hardcoded in the provider even though this is an unofficial source?

manicminer: Thanks. Unfortunately there's no general convention across apps / APIs so I elected to normalize the names so make them slightly easier to discover. Map keys without punctuation are also easier to re-use across DSLs, for example if you wanted to represent this in YAML or HCL.

t3mi: Thanks for explanation.

environments/published.go
```diff
@@ -75,6 +75,8 @@ var PublishedApis = map[string]ApiAppId{
 	"MileIqRestService":                 "b692184e-b47f-4706-b352-84b288d2d9ee",
 	"MixedReality":                      "c7ddd9b4-5172-4e28-bd29-1e0792947d18",
 	"MicrosoftAzureCli":                 "04b07795-8ddb-461a-bbee-02f9e1bf7b46",
+	"MicrosoftAzureFrontDoor"            "ad0e1c7e-6d38-4ba4-9efd-0bc77ba9f037",
```
manicminer, line 78: Syntax error

```suggestion
	"MicrosoftAzureFrontDoor":           "ad0e1c7e-6d38-4ba4-9efd-0bc77ba9f037",
```

environments/published.go
```diff
@@ -75,6 +75,8 @@ var PublishedApis = map[string]ApiAppId{
...
+	"MicrosoftAzureFrontDoorCdn"         "205478c0-bd83-4e1b-a9d6-db63a3e1e1c8",
```
manicminer, line 79: and here
```suggestion
	"MicrosoftAzureFrontDoorCdn":        "205478c0-bd83-4e1b-a9d6-db63a3e1e1c8",
```

manicminer (approved):
LGTM, just need to fix the syntax errors

t3mi:
What a failure in two strings! @manicminer sorry for that..

PR #179, opened 2022-09-21
Add support for B2CUserFlow CRUD operations

jarifibrahim:
Related to https://github.com/hashicorp/terraform-provider-azuread/issues/175
This commit adds the following
- B2CUserFlowClient for interacting with /identity/b2cuserflow endpoint
- B2CUserFlow model for the request/response payload
- Some tests for the CRUD operations


msgraph/models.go
```diff
@@ -1685,3 +1685,25 @@ type EmployeeOrgData struct {
 	CostCenter *string `json:"costCenter,omitempty"`
 	Division   *string `json:"division,omitempty"`
 }
+
+type UserFlowType int
+
+const (
+	SIGNUP_USERFLOWTYPE UserFlowType = iota
```
jarifibrahim, line 1692: Inspired by https://github.com/microsoftgraph/msgraph-beta-sdk-go/blob/f103612c4d8a3ef8e9fd15610c844381bb7b23a8/models/user_flow_type.go#L6

msgraph/models.go
```diff
@@ -1685,3 +1685,25 @@ type EmployeeOrgData struct {
...
+	SIGNIN_USERFLOWTYPE
+	SIGNUPORSIGNIN_USERFLOWTYPE
+	PASSWORDRESET_USERFLOWTYPE
+	PROFILEUPDATE_USERFLOWTYPE
+	RESOURCEOWNER_USERFLOWTYPE
+	UNKNOWNFUTUREVALUE_USERFLOWTYPE
+)
+
+type B2CUserFlow struct {
```
jarifibrahim, line 1701: B2cUserFlow has a bunch of fields https://learn.microsoft.com/en-us/graph/api/resources/b2cidentityuserflow?view=graph-rest-beta . I've added only the minimal ones. 

I can add more fields if required.

jarifibrahim:
Hello @manicminer :wave:  ! Can you please review this PR?

msgraph/models.go
manicminer: We will want to add these in time, but I'm happy to merge with these fields for now.

;qa
```diff
@@ -0,0 +1,171 @@
+package msgraph
```
manicminer, line 1: Is this file an accidental shell redirect? Could you delete this and rebase to remove it from history, thanks!

manicminer (changes requested):
@jarifibrahim Thanks for this contribution! This mostly LGTM, it just looks like an errant file made it into the commit history. If you can fix that then this should be good to merge.

;qa
jarifibrahim: Yes. It is an accidental file. Removed it in https://github.com/manicminer/hamilton/pull/179/commits/e4be3b4e541d11d14ca4a99185569f5e22731631

PR #182, opened 2022-09-27
Add Support for CRUD operations on UserFlowAttributes

jarifibrahim:
This commit adds support for CRUD operations on UserFlowAttributes.
API: https://learn.microsoft.com/en-us/graph/api/resources/identityuserflowattribute

### DIFF
--- a/msgraph/models.go
+++ b/msgraph/models.go
@@ -1687,9 +1687,9 @@ type EmployeeOrgData struct {
 }
 
 type UserFlowAttribute struct {
-	ID                    *string `json:"id,omitempty"`
-	Description           *string `json:"description,omitempty"`
-	DisplayName           *string `json:"displayName,omitempty"`
-	UserFlowAttributeType *string `json:"userFlowAttributeType,omitempty"`
-	DataType              *string `json:"dataType,omitempty"`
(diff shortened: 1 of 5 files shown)


msgraph/models.go
```diff
@@ -1685,3 +1685,11 @@ type EmployeeOrgData struct {
 	CostCenter *string `json:"costCenter,omitempty"`
 	Division   *string `json:"division,omitempty"`
 }
+
+type UserFlowAttribute struct {
+	ID                    *string `json:"id,omitempty"`
+	Description           *string `json:"description,omitempty"`
+	DisplayName           *string `json:"displayName,omitempty"`
+	UserFlowAttributeType *string `json:"userFlowAttributeType,omitempty"`
+	DataType              *string `json:"dataType,omitempty"`
```
matthiasng, line 1694: The documentation says that only "string", "boolean" and "int64" are supported.

Should we add these values here ?: https://github.com/manicminer/hamilton/blob/main/msgraph/valuetypes.go
I think this will make it a little less error prone.

jarifibrahim: Sure! Added it in https://github.com/manicminer/hamilton/pull/182/commits/44848901e799dfbfc1ebd2a3eb42bca69bbbf79f

PR #183, opened 2022-09-29
Service Principals: support for `oauth2PermissionScopes` property which is present on the v1.0 API

manicminer:


katbyte (approved):
LGTM 🍄

PR #184, opened 2022-09-30
Use Beta version for Userflow Attributes API

jarifibrahim:
The user flow attributes API was using version 1.0 but it should be beta.
Also, the URL was incorrect for the Update endpoint.

### DIFF
--- a/msgraph/userflow_attributes.go
+++ b/msgraph/userflow_attributes.go
@@ -18,7 +18,7 @@ type UserFlowAttributesClient struct {
 // NewUserFlowAttributesClient returns a new UserFlowAttributesClient.
 func NewUserFlowAttributesClient(tenantId string) *UserFlowAttributesClient {
 	return &UserFlowAttributesClient{
-		BaseClient: NewClient(Version10, tenantId),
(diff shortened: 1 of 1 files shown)


manicminer (changes requested):
Hi @jarifibrahim, thanks for opening this PR. Is there a specific issue with the v1.0 API to justify changing the default version to beta? Also, if you could rebase this onto `main` to resolve the conflicts and pull in some recent test fixes, then this should be good to merge. Thanks!

jarifibrahim:
>Hi @jarifibrahim, thanks for opening this PR. Is there a specific issue with the v1.0 API to justify changing the default version to beta? Also, if you could rebase this onto main to resolve the conflicts and pull in some recent test fixes, then this should be good to merge. Thanks!

@manicminer The userflow attribute API is available only in beta version of the graph sdk.

manicminer:
Is there a particular field that you're not able to use on the v1.0 API, because user flow attributes [seem to be available](https://learn.microsoft.com/en-us/graph/api/resources/identityuserflowattribute?view=graph-rest-1.0) and the tests are passing?

PR #187, opened 2022-10-07
quick fixes for two issues: nil pointer and type mismatch

stanleyz:
This PR fixes two issues:

1. type mis-match for "Description" field of "AccessPackageResource", reference: https://learn.microsoft.com/en-us/graph/api/resources/accesspackageresource?view=graph-rest-beta
This is introduced recently.
3. For AccessPackageResource, when there is no match on Azure, AccessPackageResources will be an empty array, hence the fix

### DIFF
--- a/msgraph/accesspackageresource.go
+++ b/msgraph/accesspackageresource.go
@@ -84,7 +84,7 @@ func (c *AccessPackageResourceClient) Get(ctx context.Context, catalogId string,
 
 	accessPackageResources := data.AccessPackageResources
 
-	if accessPackageResources == nil || len(accessPackageResources) == 0 {
+	if len(accessPackageResources) == 0 {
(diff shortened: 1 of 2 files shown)


stanleyz:
@kaovd @manicminer can you please take a look and merge this quickly? Thanks.

They are all quick simple bugs/fixes to me.

manicminer (approved):
Thanks @stanleyz for these bugfixes!

stanleyz:
Thanks @manicminer. The golang-lint workflow error here https://github.com/manicminer/hamilton/actions/runs/3238793336/jobs/5324839976 seems not related to this PR, but it does reveal an excessive check on my local, which I've fixed in the additional commit.

manicminer:
Rebased for test fixes

PR #190, opened 2022-10-09
Intune Configuration 

alexwilcox9:
Hi @manicminer 

It's been a while and I've finally gotten around to experimenting with Intune config as per  https://github.com/hashicorp/terraform-provider-azuread/issues/524

With this endpoint there's a lot of possible resource types and sub types that can nest ad infinitum. I've had a crack at mimicking the method we used on Named Locations to identify the type, unmarshal it and then return an interface.
It's proving to be a bit messy and I'm struggling to imagine how you would go about mapping something like this into a terraform resource?

I wanted to get your thoughts before really fleshing this out and sinking a lot of time into it. Thanks in advance!

### DIFF
--- a/internal/test/testing.go
+++ b/internal/test/testing.go
@@ -105,6 +105,7 @@ type Test struct {
 	ClaimsMappingPolicyClient                 *msgraph.ClaimsMappingPolicyClient
 	ConditionalAccessPoliciesClient           *msgraph.ConditionalAccessPoliciesClient
 	DelegatedPermissionGrantsClient           *msgraph.DelegatedPermissionGrantsClient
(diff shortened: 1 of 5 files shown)


soer7022:
Can I ask why this was closed? @alexwilcox9 
Very much interested in potentially using this for writing terraform providers for intune.

alexwilcox9:
@soer7022 it's definitely something I want to come back to when time and work priorities allow it.
The self-nesting nature of this makes it a particular challenge for Terraform so I may come back and try a simpler Intune endpoint

I won't be deleting any of this code though so feel free to pick it up and continue playing with it

PR #191, opened 2022-10-13
Add Transitive Member Listing for Groups

panic-kbutton:
This PR extends the SDK to include transitive listing of AAD group members, useful for retrieving nested group memberships as a flat list. The [/groups/{id}/transitiveMembers](https://learn.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0&tabs=http) endpoint is used from Microsoft Graph. This change is required to close [issue #881](https://github.com/hashicorp/terraform-provider-azuread/issues/881) within AAD's terraform provider.


manicminer (approved):
Thanks @panic-kbutton, this LGTM!

PR #194, opened 2022-11-05
use eq rather than startswith which returns a list of resources

stanleyz:
The startswith function returns a list of resources, should use eq instead, see below screenshot.

Will add test for accesspackageresource later on quite possibly in a separate PR.

accesspackageresource


stanleyz:
@manicminer can you please take a look? Thanks.

manicminer (approved):
Thanks @stanleyz, this LGTM

PR #197, opened 2022-11-29
add writeback to Group model

### LINKED ISSUE #196: Group writebackConfiguration
Hi, I have noticed that writebackConfiguration is not currently supported, is there a special reason for that? Is there another way to achieve that?

As far as I understand, all it requires is adding this to the group creation payload:
```
    "writebackConfiguration": {
        "isEnabled": true,
        "onPremisesGroupType":  UniversalSecurityGroup
    },
```

If there is no reason for this to not be there, I can make a PR for it

tomasmota:
Fixes #196 

Added ability to include writeback configuration.
Let me know if any further testing is needed.

### DIFF
--- a/msgraph/models.go
+++ b/msgraph/models.go
@@ -909,7 +909,7 @@ type Group struct {
 	Theme                         *GroupTheme                         `json:"theme,omitempty"`
 	UnseenCount                   *int                                `json:"unseenCount,omitempty"`
 	Visibility                    *GroupVisibility                    `json:"visibility,omitempty"`
-	WritebackConfiguration        *WritebackConfiguration             `json:"writebackConfiguration,omitempty"`
+	WritebackConfiguration        *GroupWritebackConfiguration        `json:"writebackConfiguration,omitempty"`
 }
 
 func (g Group) MarshalJSON() ([]byte, error) {
@@ -990,7 +990,7 @@ type GroupOnPremisesProvisioningError struct {
 	Value                *string   `json:"value,omitempty"`
 }
 
-type WritebackConfiguration struct {
+type GroupWritebackConfiguration struct {
 	IsEnabled           *bool                `json:"isEnabled"`
 	OnPremisesGroupType *OnPremisesGroupType `json:"onPremisesGroupType"`
 }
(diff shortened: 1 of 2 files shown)


msgraph/models.go
```diff
@@ -908,7 +909,7 @@ type Group struct {
 	Theme                         *GroupTheme                         `json:"theme,omitempty"`
 	UnseenCount                   *int                                `json:"unseenCount,omitempty"`
 	Visibility                    *GroupVisibility                    `json:"visibility,omitempty"`
-	IsAssignableToRole            *bool                               `json:"isAssignableToRole,omitempty"`
+	WritebackConfiguration        *WritebackConfiguration             `json:"writebackConfiguration,omitempty"`
```
manicminer, line 912: Let's name the struct `GroupWriteBackConfiguration` to match the API model

```suggestion
	WritebackConfiguration        *GroupWritebackConfiguration             `json:"writebackConfiguration,omitempty"`
```

msgraph/models.go
```diff
@@ -989,6 +990,11 @@ type GroupOnPremisesProvisioningError struct {
 	Value                *string   `json:"value,omitempty"`
 }
 
+type WritebackConfiguration struct {
```
manicminer, line 993: As above (will also need reordering)

```suggestion
type GroupWritebackConfiguration struct {
```

msgraph/valuetypes.go
```diff
@@ -448,6 +448,14 @@ const (
 	KeyCredentialUsageVerify KeyCredentialUsage = "Verify"
 )
 
+type OnPremisesGroupType = string
+
+const (
+	UniversalDistributionGroup        OnPremisesGroupType = "universalDistributionGroup"
+	UniversalSecurityGroup            OnPremisesGroupType = "universalSecurityGroup"
+	UniversalMailEnabledSecurityGroup OnPremisesGroupType = "universalMailEnabledSecurityGroup"
```
manicminer, line 456: Minor but let's order these

manicminer (approved):
@tomasmota Thanks for this PR, this is mostly looking good! I have just a few comments below, which are minor and I'll fix up shortly.

PR #198, opened 2022-11-29
Update model for DirectoryObject

MarkDordoy:
This PR adds the optional ObjectID to the DirectoryObject struct. 

This is required when using the [beta API](https://learn.microsoft.com/en-us/graph/api/applicationtemplate-instantiate?view=graph-rest-beta&tabs=http) against the application template instantiate endpoint. The response contains an Application and Service Principal which have objectId rather than Id.


tombuildsstuff (approved):

MarkDordoy:
:shipit:

PR #199, opened 2022-11-30
fix on prem group type casing

tomasmota:
Msgraph stores this as Pascal and sends it back in that way so we should have the same naming.

### DIFF
--- a/msgraph/valuetypes.go
+++ b/msgraph/valuetypes.go
@@ -451,9 +451,9 @@ const (
 type OnPremisesGroupType = string
 
 const (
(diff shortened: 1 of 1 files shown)


tomasmota:
@manicminer this is a small fix to #196

manicminer (approved):
Thanks @tomasmota, LGTM 👍

tomasmota:
Thanks @manicminer, would you mind making a new release at some point? Without this fix this PR https://github.com/hashicorp/terraform-provider-azuread/pull/964 does not work properly.

manicminer:
@tomasmota No problem, have tagged v0.53.0 with this fix

PR #204, opened 2023-01-12
Add RoleEligibilityScheduleRequestClient

JonasBak:
This PR adds a `RoleEligibilityScheduleRequestClient`, which lets us use the [PIM API for managing role eligibilities](https://learn.microsoft.com/en-us/graph/api/resources/privilegedidentitymanagementv3-overview?view=graph-rest-1.0#pim-api-for-managing-role-eligibilities). This is part of a PR to the https://github.com/hashicorp/terraform-provider-azuread repo, where we would like to be able to manage role eligibilities through terraform.

I havn't written any tests, as I wanted to get some input on how this should be done. The user/service principal requires [a role and a permission](https://learn.microsoft.com/en-us/graph/api/resources/privilegedidentitymanagementv3-overview?view=graph-rest-1.0#permissions-and-privileges) to be able to use the API. Should I set up the permissions as part of the test (and remove them after)? Should I assume the calling service principal has the correct roles and permissions?

### DIFF
--- a/msgraph/role_eligibility_schedule_request.go
+++ b/msgraph/role_eligibility_schedule_request.go
@@ -1,0 +1,120 @@
+package msgraph
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/hashicorp/go-azure-sdk/sdk/odata"
+)
+
+var actionAdminAssign string = "adminAssign"
+var actionAdminRemove string = "adminRemove"
+
+// RoleEligibilityScheduleRequestClient performs operations on RoleEligibilityScheduleRequests.
+type RoleEligibilityScheduleRequestClient struct {
+	BaseClient Client
+}
+
+// NewRoleEligibilityScheduleRequest returns a new RoleEligibilityScheduleRequestClient
+func NewRoleEligibilityScheduleRequestClient() *RoleEligibilityScheduleRequestClient {
+	return &RoleEligibilityScheduleRequestClient{
+		BaseClient: NewClient(Version10),
+	}
+}
+
+// Get retrieves a UnifiedRoleEligibilityScheduleRequest
+func (c *RoleEligibilityScheduleRequestClient) Get(ctx context.Context, id string, query odata.Query) (*UnifiedRoleEligibilityScheduleRequest, int, error) {
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
+		OData:                  query,
+		ValidStatusCodes:       []int{http.StatusOK},
+		Uri: Uri{
+			Entity: fmt.Sprintf("/roleManagement/directory/roleEligibilityScheduleRequests/%s", id),
+		},
+	})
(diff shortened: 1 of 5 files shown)


manicminer:
Hi @JonasBak, thanks for submitting this, your contribution is very appreciated! Apologies for the delayed response - this is looking good so far! When it comes to the tests, don't worry about the permissions as we'll add these as needed to our testing principal (it doesn't have the privilege to create its own role assignments anyway).

JonasBak:
I lost access to the "sandbox tenant" I used when I started working on this, so I didn't prioritize it for a while, but it looks like someone created their own PR 🙌 If that gets merged, I'll just update my PR to the terraform provider.

JonasBak:
I rebased and added a basic test

msgraph/role_eligibility_schedule_request.go
```diff
@@ -0,0 +1,120 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/hashicorp/go-azure-sdk/sdk/odata"
+)
+
+var actionAdminAssign string = "adminAssign"
+var actionAdminRemove string = "adminRemove"
```
manicminer, line 14: As per above, these values should be constants in `valuetypes.go`

msgraph/role_eligibility_schedule_request.go
manicminer, line 1: We will also want to support the `List` and `Cancel` operations for this resource

msgraph/role_eligibility_schedule_request.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	_, status, _, err = c.BaseClient.Post(ctx, PostHttpRequestInput{
+		Body:             body,
+		ValidStatusCodes: []int{http.StatusCreated},
+		Uri: Uri{
+			Entity: "/roleManagement/directory/roleEligibilityScheduleRequests",
+		},
+	})
+	if err != nil {
+		return status, fmt.Errorf("RoleAssignments.BaseClient.Get(): %v", err)
```
manicminer, line 116: We'll want to check through all the error messages to make sure they reflect the correct client and method.

manicminer (changes requested):
@JonasBak Thanks for continuing to work on this and for adding the test. This is off to a great start and with some extra work to support all the fields and methods for this resource, this will be a great addition. I've made some comments below - if you can take a look at these, I'll be happy to re-review and we can look to try and get this merged. Thanks!

JonasBak:
Thanks for the feedback 🙌 I have some problems with the test being flaky, as it sometimes fails with `RoleNotFound: The role is not found`, what can I do to improve this?

JonasBak:
Also, should I add the `recurrence` field in [requestSchedule](https://learn.microsoft.com/en-us/graph/api/resources/requestschedule?view=graph-rest-1.0)? It says that it's unsupported in PIM.

manicminer (approved):
Thanks again for your work on this @JonasBak. This is looking good to me, I think it's fine to omit the test for the `Cancel` method for now, we can look at it again once we can support the entire workflow.

The test failure seems unrelated, so I'm happy to merge this. Thanks again!

PR #206, opened 2023-01-18
Provide a function to create a group directly into an administrative unit

SwissGipfel:
We'd like to provide an additional function on the AdministrativeUnitsClient which allows us to create group objects directly into an administrative unit, as documented on [MS Docs](https://learn.microsoft.com/en-us/azure/active-directory/roles/admin-units-members-add#create-a-new-group-in-an-administrative-unit-2).

This would enable managing group in an administrative unit without write permission on a whole tenant. As it states [here](https://learn.microsoft.com/en-us/azure/active-directory/roles/admin-units-members-add#prerequisites), the permissions could be scoped only to the respective AU.
It would also enable to solve [this issue](https://github.com/hashicorp/terraform-provider-azuread/issues/906) on the terraform provider for azuread, as the provider uses this package as dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.53.0 (Unreleased)
 
+- Add the App ID for `Microsoft.StorageSync` to `PublishedApis` ([#200](https://github.com/manicminer/hamilton/pull/200))
 - Bugfix: Fix casing of values for `OnPremisesGroupType` type ([#199](https://github.com/manicminer/hamilton/pull/199))
 
 ## 0.52.0 (November 30, 2022)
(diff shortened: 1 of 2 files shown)


ccadruvi:
Hey @manicminer, we couldn't reproduce the test failure, do you think the tests fail because of this addition?

manicminer:
> Hey @manicminer, we couldn't reproduce the test failure, do you think the tests fail because of this addition?

I think it was just some missing roles/permissions on our end, I think I have fixed it now

manicminer (approved):
@SwissGipfel Thanks for this contribution! This LGTM 👍

PR #207, opened 2023-01-18
Extend Conditional access session controls  model by new parameter disableResilienceDefaults

stawik-mesa:
Add new field to model.

Will be used later in https://github.com/hashicorp/terraform-provider-azuread to fix https://github.com/hashicorp/terraform-provider-azuread/issues/931


manicminer (approved):
Hi @stawik-mesa, thanks for this PR and apologies for the delay in reviewing. This looks good to me 👍 

EDIT: I only just noticed the test failure - it looks like `disableResilienceDefaults` is a simple boolean in both API versions, so I've updated the model accordingly.

PR #208, opened 2023-01-19
Add entitlement management role clients

alexwilcox9:
Hi

This PR adds two new clients `EntitlementRoleAssignmentsClient` and `EntitlementRoleDefinitionsClient` based off of `RoleAssignmentsClient` and `RoleDefinitionsClient`
The only notable difference is that custom roles are not supported so the `EntitlementRoleDefinitionsClient` does not have any create/update/delete functions

https://learn.microsoft.com/en-us/graph/api/rbacapplication-post-roleassignments


manicminer (approved):
Thanks for picking this up @alexwilcox9, this LGTM! 👍

PR #209, opened 2023-01-22
Continuing Terms of Use

alexwilcox9:
Picking up where https://github.com/manicminer/hamilton/pull/62 left off

I had a go at using the new `file` property but didn't have much luck. If both `Files` and `File` are empty the response is:
`unexpected status 400 with OData error: BadRequest: Either policyFileLocalizations or policyFile has to be populated.`
However trying to send only `File` results in:
`unexpected status 400 with OData error: BadRequest: Value cannot be null. Parameter name: source`

Googling this the only related article I could find was https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/652 and as far as I can see they are using `Files` which does work for us
```
Running tool: /home/linuxbrew/.linuxbrew/bin/go test -timeout 10m -run ^TestTermsOfUseAgreementClient$ github.com/manicminer/hamilton/msgraph

ok  	github.com/manicminer/hamilton/msgraph	5.094s
```

It's probably worth adding a client for creating agreements on existing ToU as shown here:
https://learn.microsoft.com/en-us/graph/api/agreement-post-files

Let me know what you think

Thanks


manicminer (approved):
Thanks for this @alexwilcox9, appreciate your tracking this and figuring out the endpoint bugs. LGTM 👍

PR #210, opened 2023-01-24
Add support for entitlement management access package assignment requests

MarkDordoy:
Adds support for MS Graph Entitlement management access package assignment requests. 

This PR creates the required client and the following methods:

* [Create](https://learn.microsoft.com/en-us/graph/api/entitlementmanagement-post-assignmentrequests?view=graph-rest-1.0&tabs=http)
* [Get](https://learn.microsoft.com/en-us/graph/api/accesspackageassignmentrequest-get?view=graph-rest-1.0&tabs=http)
* [List](https://learn.microsoft.com/en-us/graph/api/entitlementmanagement-list-assignmentrequests?view=graph-rest-1.0&tabs=http)
* [Cancel](https://learn.microsoft.com/en-us/graph/api/accesspackageassignmentrequest-cancel?view=graph-rest-1.0&tabs=http)
* [Reprocess](https://learn.microsoft.com/en-us/graph/api/accesspackageassignmentrequest-reprocess?view=graph-rest-1.0)
* [Delete](https://learn.microsoft.com/en-us/graph/api/accesspackageassignmentrequest-delete?view=graph-rest-1.0&tabs=http)

I've created the model so it supports both [V1.0](https://learn.microsoft.com/en-us/graph/api/resources/accesspackageassignmentrequest?view=graph-rest-1.0) and [beta](https://learn.microsoft.com/en-us/graph/api/resources/accesspackageassignmentrequest?view=graph-rest-beta)

I've updated the model for `AccessPackage` to now support the Relationships that can be retrieved using the odata Expand.Relationship

I've added tests where possible. I'm unable to test Reprocess as this can only be used when azure has a failure which we cannot simulate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 0.54.0 (Unreleased)
+## 0.55.0 (January 27, 2023)
+
+- Support for [creating groups in an administrative unit](https://learn.microsoft.com/en-us/azure/active-directory/roles/admin-units-members-add#create-a-new-group-in-an-administrative-unit-2) ([#206](https://github.com/manicminer/hamilton/pull/206))
+
+## 0.54.0 (January 13, 2023)
 
 - Support for `Description` field in the `Application` model ([#205](https://github.com/manicminer/hamilton/pull/205))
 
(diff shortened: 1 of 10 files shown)


manicminer (approved):
Thanks @MarkDordoy, this looks great to me 👍

PR #211, opened 2023-02-03
Change PreferredTokenSigningKeyThumbprint to nullable

tagur87:
When working on implementing a resource to set the PreferredTokenSigningKeyThumbprint, it was found that this field needs to be able to be set to null if the field is empty.

This is required to remove a preferred token if desired.


tagur87:
@manicminer - found this bug when working on the new resource. Seems this is needed to allow the field to be set to null.

Thanks!

manicminer (approved):
Thanks for this fix @tagur87, this LGTM 👍

PR #212, opened 2023-02-04
Add support for pkcs12 bundles

kiall:
Switch to using `software.sslmate.com/src/go-pkcs12`'s DecodeChain method in order to allow loading bundles that contain a full cert chain.


manicminer:
(Test failure unrelated and due to PR from fork)

PR #213, opened 2023-02-08
Auth: Add supports for custom command

magodo:
Custom command allows users to specify custom command that is used to retrieve the access token.

This covers cases like:

- Azure CLI delegation
- Azure PWSH delegation
- Accept plain access token


manicminer:
Hi @magodo, thanks for suggesting this. I think this sort of custom authorizer is best left to users to implement, which they can do using the auth.Authorizer interface. In terms of supporting this sort of authorizer in the Azure providers, as mentioned internally this is unlikely to be added due to the issues of renewability and obtaining tokens for each and every data plane API that might be used which is highly impractical. Accordingly, I don't expect we'll merge this and so I'm going to close this one for now.

magodo:
@manicminer With this implementation, I think both the expiry and different endpoint (incl data plane API) problems are solved:

- On expiry, the `TokenSource`, i.e. the `CustomCommandAuthorizer` will be called by its `Token()` method, in which case, the command (e.g. az cli) is responsible to provide an up-to-date token. This is also how the `AzureCliAuthorizer` works
- Different endpoint, as each authorizer is built for the input endpoint (i.e. `environment.Api`)

My motivation is to ultimately make the azure provider supports this functionality. For azurerm, it is fine to implement this in the `hashicorp/go-azure-helper` module. But azuread doesn't depend on that, which makes me think this is the right place to add this support. WDYT?

manicminer:
@magodo The AzureAD and AzureRM will be converging on the same auth library in the near future, so that will help in terms of consolidating features like this.

For Powershell support, I would prefer to support this explicitly, as we do for Azure CLI, rather than have an open ended mechanism for plugging in any binary - this would be very difficult to support and I am not comfortable with the security implications this would have for users.

PR #215, opened 2023-02-10
Add TokenIssuancePolicy

kenchan0130:
This adds initial support for the TokenIssuancePolicy resource.

Represents the policy to specify the characteristics of SAML tokens issued by Azure AD. You can use token issuance policies to:

* Set signing options
* Set signing algorithm
* Set SAML token version

I believe it will also be useful for adding resources to azure ad terraform in the future, as it will be used to configure the SAML signature algorithm, etc.

https://learn.microsoft.com/en-us/graph/api/resources/tokenissuancepolicy?view=graph-rest-1.0


msgraph/applications.go
```diff
@@ -763,3 +763,132 @@ func (c *ApplicationsClient) DeleteFederatedIdentityCredential(ctx context.Conte
 
 	return status, nil
 }
+
+// AssignTokenIssuancePolicy assigns a tokenIssuancePolicy to an application
+func (c *ApplicationsClient) AssignTokenIssuancePolicy(ctx context.Context, application *Application) (int, error) {
```
manicminer, line 768: To make this more consistent with other child objects, could we make change this to accept an `applicationId` and a `TokenIssuancePolicy` (e.g. similar to `CreateFederatedIdentityCredential` or `AddPassword`)?

manicminer (changes requested):
Hi @kenchan0130, thanks for contributing to this SDK! Overall this is looking great. I have just one comment about a method signature/behavior which is mainly about trying to be consistent between client methods. If you can take a look at this, this should be good to merge. Thanks!

msgraph/applications.go
kenchan0130: Thanks for the review.
Which of the following interfaces do you prefer?
I am willing to adapt to your ideas.

```go
// pattern 1
AssignTokenIssuancePolicy(ctx context.Context, applicationId string, policy TokenIssuancePolicy)
```

```go
// pattern 2
AssignTokenIssuancePolicy(ctx context.Context,  applicationId string, policies *[]TokenIssuancePolicy)
```

kenchan0130: @manicminer oops. I forgot to mention.

manicminer: @kenchan0130 Apologies for delayed reply, the second form should work well I think?

kenchan0130: @manicminer 
I don't even know why I wrote codes assigning  TokenIssuancePolicy to the application.
TokenIssuancePolicy should have been tied to the service principal, not the application.
Therefore, I have changed like https://github.com/manicminer/hamilton/pull/215/commits/bbf153dac0f9a28795b4bfda433224f11fe68964.

msgraph/serviceprincipals.go
```diff
@@ -778,3 +778,135 @@ func (c *ServicePrincipalsClient) AssignAppRoleForResource(ctx context.Context,
 
 	return &appRoleAssignment, status, nil
 }
+
+// AssignTokenIssuancePolicy assigns tokenIssuancePolicies to a service principal
+func (c *ServicePrincipalsClient) AssignTokenIssuancePolicy(ctx context.Context, servicePrincipal *ServicePrincipal) (int, error) {
```
kenchan0130, line 783: @manicminer 
As you have said in https://github.com/manicminer/hamilton/pull/215#discussion_r1104378909, I have made the interface like `AssignClaimsMappingPolicy`.

However, I would like to be sure that my current changes are not a problem for the overall design policy of `ServicePrincipalClient`, as it differs from [your opinion](https://github.com/manicminer/hamilton/pull/215#discussion_r1112971782).

kenchan0130:
The fail in [this test](https://github.com/manicminer/hamilton/actions/runs/4234570900/jobs/7357215852) does not appear to be related to any change in my code.

```log
=== RUN   TestConnectedOrganizationClient
    connectedorganization_test.go:68: ConnectedOrganizationClient.Create(): ConnectedOrganizationClient.BaseClient.Post(): unexpected status 400 with OData error: ConnectedOrganizationAlreadyExists: The connected organziation already exists.
--- FAIL: TestConnectedOrganizationClient (0.45s)
```

msgraph/serviceprincipals.go
kenchan0130: @manicminer 

> Thanks for updating. As per my earlier comment, I think it would be best to use a function signature like

OK, I understand.

> Secondly, for the question of whether this operation should happen for applications or service principals, I note that the docs only seem to cover the `/applications/{id}/tokenIssuancePolicies` endpoint - could it be that these are applicable to both `/applications/{id}/tokenIssuancePolicies and /servicePrincipals/{id}/tokenIssuancePolicies`?

I have tried to apply a policy to the application as follows.

```sh
curl 'https://graph.microsoft.com/v1.0/applications/a191c3d6-9371-446a-a4aa-647226be2f1b/tokenIssuancePolicies/$ref' \
  -H 'Authorization: Bearer xxxxx' \
  --data-raw '{"@odata.id":"https://graph.microsoft.com/v1.0/policies/tokenIssuancePolicies/8efb6001-1369-4c68-be2d-04182c8edd76"}'
```

```json
{
    "error": {
        "code": "Request_BadRequest",
        "message": "Policy operations on v2 application are disabled.",
        "innerError": {
            "date": "2023-03-02T14:55:42",
            "request-id": "3ce11a1a-d63a-4100-9bf8-c83ca1666fbd",
            "client-request-id": "663c72ad-ccc3-2eef-38aa-3ea17b11a44b"
        }
    }
}
```

I have found that this request results in the error.
Therefore, I chose not to implement it for the application.

kenchan0130:
@manicminer 
I have responded to your change request.
Please review my code.

kenchan0130:
@manicminer How has the progress been coming along?

manicminer (approved):
@kenchan0130 Many thanks for clarifying and updating. Sorry for the delay, this LGTM 🚀

Issue #169, opened 2022-07-20
Support Additional Azure Environments


blheatwole:
[EnvironmentFromString](https://github.com/manicminer/hamilton/blob/82c4c5a2f374f826d0c49b0a839bebbe5009d49c/environments/environments.go#L103) hardcodes a list of supported environments. The Hashicorp Terraform Provider for AzureRM makes use of EnvironmentFromString [here](https://github.com/hashicorp/terraform-provider-azurerm/blob/852010ec4bf8fcf32e00051138de124032f6022f/internal/clients/builder.go#L61)

Because the list is hardcoded, the terraform provider is useless for [Azure environments whose URLs cannot be publicly published](https://azure.microsoft.com/en-us/global-infrastructure/government/national-security/). Azure provides a method to discover the URLs for all the endpoints referenced in the Environment structure by using the metadata URL. [A sample implementation of using this URL is located here](https://github.com/Azure/go-autorest/blob/7dd32b67be4e6c9386b9ba7b1c44a51263f05270/autorest/azure/metadata_environment.go#L96)

Please add `EnvironmentFromMetadataUrl(url string) (Environment, error)` to support additional Azure environments and make hamilton more globally useful.

Thanks!

manicminer:
Hi @blheatwole, thanks for requesting. We have actually previously implemented a function to look up the MS Graph URL from an ARM metadata service - however to my knowledge MS Graph endpoints are still not published this way, making it impossible for us to support private/custom clouds.

I'm happy to leave this issue open to track this, although it is blocked at present until someone can demonstrate that an MS Graph endpoint is being published in a custom cloud. Ideally, it should be published for all clouds.

blheatwole:
Running `curl https://management.azure.com/metadata/endpoints?api-version=latest` returns the following JSON:
```
[
  {
    "portal": "https://portal.azure.com",
    "authentication": {
      "loginEndpoint": "https://login.microsoftonline.com/",
      "audiences": [
        "https://management.core.windows.net/",
        "https://management.azure.com/"
      ],
      "tenant": "common",
      "identityProvider": "AAD"
    },
    "media": "https://rest.media.azure.net",
    "graphAudience": "https://graph.windows.net/",
    "graph": "https://graph.windows.net/",
    "name": "AzureCloud",
    "suffixes": {
      "azureDataLakeStoreFileSystem": "azuredatalakestore.net",
      "acrLoginServer": "azurecr.io",
[...]

PR #217, opened 2023-02-13
Migrate auth and environments packages to hashicorp/go-azure-sdk

manicminer:
The `auth` and `environments` packages in the `hashicorp/go-azure-sdk` module now have feature parity with the packages here, so this migrates the SDK to these newer packages and removes the `auth` and `environments` packages from this module.

This is a breaking change, although with minimal disruption as can be seen from the changes to the test helpers in this PR.

Resolves: #169 

Depends on: https://github.com/hashicorp/go-azure-sdk/pull/306


go.mod
```diff
@@ -4,11 +4,12 @@ go 1.16
 
 require (
 	github.com/golang/protobuf v1.5.2 // indirect
+	github.com/hashicorp/go-azure-sdk v0.20230214.1122757-0.20230214215836-94fd7f7d8e2a
```
tombuildsstuff, line 7: presumably wants a stable release?

tombuildsstuff (approved):
One comment but otherwise 👍

PR #220, opened 2023-02-15
List group memberships for administrative unit

ccadruvi:
List administrative unit memberships for a group

Relates to https://github.com/hashicorp/terraform-provider-azuread/pull/984


manicminer (approved):
Thanks @SwissGipfel, this LGTM!

(test failures unrelated)

PR #227, opened 2023-02-23
Add User Thumbnail Upload

Threpio:
Required for: https://github.com/hashicorp/terraform-provider-azuread/issues/1012

Do I need to add a different logo than the one that is used for Applications thumbnails? No I will be the application.


msgraph/users.go
```diff
@@ -437,3 +437,23 @@ func (c *UsersClient) DeleteManager(ctx context.Context, id string) (int, error)
 
 	return status, nil
 }
+
+// UploadThumbnailPhoto uploads a thumbnail photo for the specified user which should be a gif, jpeg or png image.
+func (c *UsersClient) UploadThumbnailPhoto(ctx context.Context, userId, contentType string, thumbnailData []byte) (int, error) {
+	var status int
```
rgl, line 443: I think this line can be removed. as `status` is also defined bellow.

Threpio: The syntax and format was largely copied/rewritten from: 

https://github.com/manicminer/hamilton/blob/c99311563ba9c4bf460ff28609eb303769d20d00/msgraph/applications.go#L608

If that has the status initialized before that function then my guess is the HTTP call can sometimes return no error but a function. Am I wrong in how that would work?

rgl: The `:=` operator creates the required variables.

That `var status int` declaration is not required.

If you look [at the next function, at the file you've copied it from](https://github.com/manicminer/hamilton/blob/c99311563ba9c4bf460ff28609eb303769d20d00/msgraph/applications.go#L629), you'll see it does not uses that declaration.

Threpio: I figured that `var status int` was something I didn't understand about the HTTP returns etc. Must be legacy code.

I will remove for this function.

manicminer (approved):
Thanks @Threpio, this LGTM 👍

manicminer:
@Threpio Could you take a look at the test failure?

Threpio:
This error is bizarre - The documentation for this endpoint is duplicated between `/user/{id}/thumbnailPhoto` and `/user/{id}/photo/$value` - I will change it to use the second endpoint as that is more in line with the docs.

HOWEVER - In doing so:

https://github.com/manicminer/hamilton/blob/6b5e7f6ac1345d94d93fa890f325ae780ca6b4d2/msgraph/client.go#L448

This will need to change as the content-type header needs to be able to be set to an image format. I don't know how to determine what this will break?

manicminer (approved):
Apologies @Threpio, this slipped through amongst some unfinished PRs. I changed it from a PATCH to a PUT operation and it seems to work, this LGTM 👍

manicminer:
(Test failures unrelated)

PR #229, opened 2023-02-23
Fix some files fomart

kenchan0130:
I perform formatting when I contribute to this project.
However, each time the formatting is applied to a file that has nothing to do with my changes.
So once I ran `make fmt`.


manicminer:
Thanks for this @kenchan0130. It looks like these already got fixed in main, so I'll close this one out for now.

PR #237, opened 2023-03-30
Update model SignInActivity to support non-interactive signin

MarkDordoy:
Extends model to match documentation: https://learn.microsoft.com/en-us/graph/api/resources/signinactivity?view=graph-rest-beta 

I also fixed the name locations test so checks go green.


manicminer (approved):
Thank @MarkDordoy, this LGTM 👍

PR #240, opened 2023-04-08
Add RoleEligibilityScheduleRequest

kenchan0130:
This adds initial support for the RoleEligibilityScheduleRequest resource.

This client requires `RoleManagement.ReadWrite.Directory` permissions.

In https://github.com/manicminer/hamilton/pull/204, someone else has submitted a Pull Request, but it doesn't seem to be active. Therefore, I decided to create a new Pull Request.

Graph API Reference:

* https://learn.microsoft.com/en-us/graph/api/rbacapplication-list-roleeligibilityschedulerequests
* https://learn.microsoft.com/en-us/graph/api/rbacapplication-post-roleeligibilityschedulerequests
* https://learn.microsoft.com/en-us/graph/api/unifiedroleeligibilityschedulerequest-get

### DIFF
--- a/msgraph/role_eligibility_schedule_requests_test.go
+++ b/msgraph/role_eligibility_schedule_requests_test.go
@@ -1,0 +1,92 @@
+package msgraph_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/hashicorp/go-azure-sdk/sdk/odata"
+	"github.com/manicminer/hamilton/internal/test"
+	"github.com/manicminer/hamilton/internal/utils"
+	"github.com/manicminer/hamilton/msgraph"
+)
+
+func TestRoleEligibilityScheduleRequestsClient(t *testing.T) {
+	c := test.NewTest(t)
+	defer c.CancelFunc()
+
+	testUser := testUser_Create(t, c)
+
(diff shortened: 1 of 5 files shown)


kenchan0130:
I had to skip the implementation of https://learn.microsoft.com/en-us/graph/api/unifiedroleeligibilityschedulerequest-cancel API because it requires the status to be set to `Granted`. However, I could not find a programmable way to set the status to `Granted`.

msgraph/role_eligibility_schedule_requests_test.go
```diff
@@ -0,0 +1,92 @@
...
+			Expiration: &msgraph.ExpirationPattern{
+				Type: utils.StringPtr(msgraph.ExpirationPatternTypeNoExpiration),
+			},
+		},
+	})
+	testRoleEligibilityScheduleRequestsClient_List(t, c)
+	testRoleEligibilityScheduleRequestsClient_Get(t, c, *roleEligibilityScheduleRequest.ID)
+
+	// Processing of the request action takes a little time internally, so it is necessary to sleep for another action
+	time.Sleep(20 * time.Second)
```
kenchan0130, line 35: The `time.Sleep` function is not working as expected, causing the test to fail. This issue is also happening in my local environment. 
I have confirmed that it works correctly when I insert an output process like `fmt.Println("Debug")` before this operation in my local environment as a trial.

Since I am not familiar with Go, if anyone knows the cause of this issue, please let me know.

PR #245, opened 2023-06-15
Add delete function for `AccessPackageResourceRoleScope`

alexwilcox9:
Hey @manicminer 

Hoping this can help towards resolving https://github.com/hashicorp/terraform-provider-azuread/issues/1091
Once the delete function is merged it can be referenced in this function:
https://github.com/hashicorp/terraform-provider-azuread/blob/main/internal/services/identitygovernance/access_package_resource_package_association_resource.go#L146-L150

Errors should still be expected if a user has been assigned a given resource but this will at least allow things to delete when no assignments are in place

Let me know if any tweaks are needed!

### DIFF
--- a/msgraph/accesspackageresource.go
+++ b/msgraph/accesspackageresource.go
@@ -83,7 +83,7 @@ func (c *AccessPackageResourceClient) Get(ctx context.Context, catalogId string,
 	accessPackageResources := data.AccessPackageResources
 
 	if len(accessPackageResources) == 0 {
-		return nil, status, fmt.Errorf("No accessPackageResource found with catalogId %v and originId %v", catalogId, originId)
+		return nil, http.StatusNotFound, fmt.Errorf("No accessPackageResource found with catalogId %v and originId %v", catalogId, originId)
 	}
 
 	return &accessPackageResources[0], status, nil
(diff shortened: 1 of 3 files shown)


alexwilcox9:
The second commit addresses a bug I've noticed where if a `azuread_access_package_resource_package_association` is deleted by a user in the console terraform errors rather than seeing the resource has gone and removing it from state

This happens because the `AccessPackageResourceRoleScopeClient.Get` function returns the status of the HTTP request to the `accessPackages` endpoint which does succeed but then as it parses the output and fails to find a matching entry it does not update that status to 404
The above results in this if statement never firing
https://github.com/hashicorp/terraform-provider-azuread/blob/main/internal/services/identitygovernance/access_package_resource_package_association_resource.go#L124-L127

If it is more appropriate to always hand back the status code from the endpoint and instead this should be handled in the Terraform code let me know and I'll role this commit back

manicminer (approved):
Thanks @alexwilcox9, this LGTM.

Great spot on the missing resource bug. I think it's fine to fake a 404 error for now, this will be the least painful fix. We'll likely do it slightly differently in future but this works for now 👍

PR #246, opened 2023-07-08
Added `ServicePrincipalRiskLevels` to `ConditionalAccessConditionSet`

bubbletroubles:
Added `ServicePrincipalRiskLevels` to `ConditionalAccessConditionSet`

Created a new value type for `ConditionalAccessServicePrincipalRiskLevel` as Service Principles don't have the `hidden` value which is currently in `ConditionalAccessRiskLevel` - see [Graph API v1.0 documentation](https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessconditionset?view=graph-rest-1.0)

### DIFF
--- a/msgraph/models.go
+++ b/msgraph/models.go
@@ -615,17 +615,17 @@ type ConditionalAccessClientApplications struct {
 }
 
 type ConditionalAccessConditionSet struct {
-	Applications               *ConditionalAccessApplications                `json:"applications,omitempty"`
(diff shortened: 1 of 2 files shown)


manicminer (approved):
Thanks for this @bubbletroubles. Although the `servicePrincipalRiskLevels` field doesn't support that value, it is still the same object so we should use the `ConditionalAccessRiskLevel   ` type. I'll go ahead and update this so we can merge in the additional field.

PR #248, opened 2023-07-18
Add support for Privileged Identity Management Groups

### LINKED ISSUE #117: Support for Privilege Access Group
We would like to use terraform to allow JIT groups to get conditional access, like effective / or active.

oWretch:
As a prerequisite to addressing https://github.com/hashicorp/terraform-provider-azuread/issues/68.

I think this will (once completed) supersede #240 and close #117.


manicminer (commented):
Thanks for working on this @oWretch. I think you're right, this will likely need to be merged together. This is looking great so far. I'm not sure if you have any additional methods to add, but once we have some basic tests covering these, we can look at merging this!

oWretch (commented):
I found the root of the problem. Hadn't fully added the client to the testing package

oWretch:
I've work out that I can split this into multiple PRs, so I'll start raising those separately and link them here.

PR #250, opened 2023-07-25
Bugfix: enable `grantControls` or `sessionControls`

manicminer:
Conditional Access Policies can have `grantControls` or `sessionControls`, or both, but must be explicitly null in the `or` case


katbyte (approved):
LGTM 🌵

PR #251, opened 2023-07-26
Support `Manager` value for AccessReviewReviewerType

manicminer:
Related: https://github.com/hashicorp/terraform-provider-azuread/issues/1149


katbyte (approved):
LGTM aside from test failure 💾

manicminer:
Test failure looks to be a race condition with auto-approvals on access package assignment requests, unrelated to this change. Will fix separately due to complexity.

PR #253, opened 2023-08-11
Add support for PIM AAD Group assignments

oWretch:
This is the first PR to provide support for Privileged Management of AAD Groups. This implements part of #248. Depends on #276.

### DIFF
--- a/internal/test/testing.go
+++ b/internal/test/testing.go
@@ -91,52 +91,53 @@ type Test struct {
 	Claims *claims.Claims
 	Token  *oauth2.Token
 
-	AccessPackageAssignmentPolicyClient       *msgraph.AccessPackageAssignmentPolicyClient
-	AccessPackageAssignmentRequestClient      *msgraph.AccessPackageAssignmentRequestClient
-	AccessPackageCatalogClient                *msgraph.AccessPackageCatalogClient
-	AccessPackageClient                       *msgraph.AccessPackageClient
-	AccessPackageResourceClient               *msgraph.AccessPackageResourceClient
-	AccessPackageResourceRequestClient        *msgraph.AccessPackageResourceRequestClient
-	AccessPackageResourceRoleScopeClient      *msgraph.AccessPackageResourceRoleScopeClient
-	AdministrativeUnitsClient                 *msgraph.AdministrativeUnitsClient
-	ApplicationTemplatesClient                *msgraph.ApplicationTemplatesClient
-	ApplicationsClient                        *msgraph.ApplicationsClient
-	AppRoleAssignedToClient                   *msgraph.AppRoleAssignedToClient
-	AuthenticationMethodsClient               *msgraph.AuthenticationMethodsClient
-	AuthenticationStrengthPoliciesClient      *msgraph.AuthenticationStrengthPoliciesClient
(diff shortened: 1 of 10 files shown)


oWretch:
I haven't worked out how to run the tests under a delegated token, so I've added an environment variable with default in the test file to use a pre-existing group. Happy for assistance on getting the group activated for PIM under a delegated token.

manicminer (commented):
Hi @oWretch, thank you for submitting this! Overall this is looking really great.

For the delegated auth, we don't have any built-in support in the auth package right now, as such implementations are largely application specific and so are left to the application to implement. However, you could use the Azure CLI authentication to authenticate as a user, which would then obtain a delegated token for you, e.g. `az login --username user@domain.com --password sUpErSeCrEt`. I believe that for a noninteractive flow you're going to need no MFA enforcement for that user. I'll look at adding support for this to the GitHub runner, if you can look at configuring the tests to use CLI auth?

manicminer:
@oWretch I can't seem to find an endpoint in MS Graph for registering groups for PIM. The portal uses the PIM data plane, and the closest thing I can find in MS Graph docs is for PIMv2, and seems to be [only for Azure resources](https://learn.microsoft.com/en-us/graph/api/governanceresource-register?view=graph-rest-beta). Any ideas?

oWretch:
There isn't an API for activating in v3 - it is all meant to be seamless and happen when you use one of the standard APIs for PIM. The issue is that, for some reason, service principals can't enable a group for PIM (see the [v2 documentation](https://learn.microsoft.com/en-us/graph/api/governanceresource-register?view=graph-rest-beta#permissions)).

When I manually enable a group for PIM in the Azure AD portal, it uses an undocumented API to perform the registration, which we can't do outside of the portal (and therefore can't use a service principal).

This is why I defaulted to testing on a group that was already enabled for PIM. It seems to be a limitation of the Microsoft APIs. The only other option is to get a delegated token (as you have suggested) from a user instead of principal to do the testing. I just haven't had any time recently to look further into this.

manicminer:
@oWretch Thanks, I'll give that a try and see how far I get! We should be able to support working with a user principal for specific tests, we already have prior works for this due to other APIs also failing to support being invoked by service principals. Whilst we _could_ revert to using a pre-existing group, I'd much prefer not to as it would interfere with other tooling that we have.

oWretch:
@manicminer Did you make any progress on looking into how best to do the tests here? I may have some more time in the next couple of weeks to look into it a bit further.

oWretch:
@manicminer A service principal can now activate the PIM groups. I've updated the tests to create the group, and they run successfully under the SP. It just needs to have the following Graph permissions:
* `PrivilegedAccess.ReadWrite.AzureADGroup`
* `PrivilegedAssignmentSchedule.ReadWrite.AzureADGroup`
* `PrivilegedEligibilitySchedule.ReadWrite.AzureADGroup`

Test passes happily 🙂
```
=== RUN   TestPrivilegedAccessGroupAssignmentScheduleClient
--- PASS: TestPrivilegedAccessGroupAssignmentScheduleClient (34.19s)
PASS
ok  	github.com/manicminer/hamilton/msgraph	34.939s
```

oWretch:
Replaced by #277

PR #257, opened 2023-08-14
Add `updateAllowedCombinations` to `AuthenticationStrengthPoliciesClient`

alexwilcox9:
It turns out you can't update the `allowedCombinations` field using the normal update endpoint. You'll get the following error
```
AuthenticationStrengthPoliciesClient.BaseClient.Patch(): unexpected status 405 with OData error: methodNotAllowed: Combinations can only be updated using updateAllowedCombinations action.: Combinations can only be updated
│ using updateAllowedCombinations action.
```
Instead there is a specific endpoint you must use documented here:
https://learn.microsoft.com/en-us/graph/api/authenticationstrengthpolicy-updateallowedcombinations?view=graph-rest-1.0&tabs=http

Currently I'm passing through the whole `AuthenticationStrengthPolicy` object but technically it should only be the `allowedCombinations` field without the display name etc.
If it's preferred I can create an object that only has `allowedCombinations` for use here but wasn't sure if that was needed given that Graph seems to ignore the extra fields

### DIFF
--- a/msgraph/authentication_strength_policy.go
+++ b/msgraph/authentication_strength_policy.go
@@ -141,6 +141,42 @@ func (c *AuthenticationStrengthPoliciesClient) Update(ctx context.Context, Authe
 	return status, nil
 }
 
+// Update amends an existing AuthenticationStrengthPolicy's allowed combinations
+func (c *AuthenticationStrengthPoliciesClient) UpdateAllowedCombinations(ctx context.Context, policy AuthenticationStrengthPolicy) (int, error) {
+	var status int
+
+	if policy.ID == nil {
+		return status, errors.New("cannot update AuthenticationStrengthPolicy with nil ID")
+	}
+
+	if policy.AllowedCombinations == nil {
+		return status, errors.New("cannot update AuthenticationStrengthPolicy with nil AllowedCombinations")
+	}
+
+	allowedCombinations := AuthenticationStrengthPolicy{
+		AllowedCombinations: policy.AllowedCombinations,
+	}
+
+	body, err := json.Marshal(allowedCombinations)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


msgraph/authentication_strength_policy.go
```diff
@@ -141,6 +141,38 @@ func (c *AuthenticationStrengthPoliciesClient) Update(ctx context.Context, Authe
 	return status, nil
 }
 
+// Update amends an existing AuthenticationStrengthPolicy's allowed combinations
+func (c *AuthenticationStrengthPoliciesClient) UpdateAllowedCombinations(ctx context.Context, Policy AuthenticationStrengthPolicy) (int, error) {
```
manicminer, line 145: Can we camelcase this local variable?

```suggestion
func (c *AuthenticationStrengthPoliciesClient) UpdateAllowedCombinations(ctx context.Context, policy AuthenticationStrengthPolicy) (int, error) {
```

manicminer (changes requested):
Hi @alexwilcox9, thanks for this. I think we should build a new struct with the appropriate fields, since the API could at any time start throwing errors with the extra fields. We can still receive an `AuthenticationStrengthPolicy` as a method argument, and just pull out the `ID` and `AllowedCombinations` fileds. Other than that, this looks great 👍

alexwilcox9:
Thanks for the feedback @manicminer, I've updated the code to create a new `AuthenticationStrengthPolicy` with only the `allowedCombinations` field set.

Is this sufficient or should I create a new struct in models.go for use here?

manicminer (approved):
@alexwilcox9 Sorry for the delay, I think this is fine to construct a request like this and we don't need to define a new exported struct type for it. Thanks for this!

PR #259, opened 2023-09-25
Service Principals: support the `appMetadata` field

manicminer:
Support the `appMetadata` field, which reveals details of the managed identity the SP is linked to


jackofallops (approved):
LGTM 👍

PR #260, opened 2023-10-18
ApplicationsClient: add the `SetFallbackPublicClient()` method

manicminer:
Also update fields to `StringNullWhenEmpty` in `InformationalUrl` model


manicminer:
Test failure is unrelated

Issue #261, opened 2023-10-23
Add support for conditional access policy guestsOrExternalUsers settings


agileknight:
Support `conditions.users.includeGuestsOrExternalUsers` and `conditions.users.excludeGuestsOrExternalUsers` fields on conditional access policy.

See
* https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessusers?view=graph-rest-1.0
* https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessguestsorexternalusers?view=graph-rest-1.0

PR #264, opened 2023-10-26
Fix missing odata type for externalTenants configuration

agileknight:
Without setting the correct @odata.type for both
possible values all and enumerated, a patch will fail when first enumerating and setting members and then changing to type all because the stored object
is of the wrong type.

References #261

### DIFF
--- a/msgraph/conditionalaccesspolicy_test.go
+++ b/msgraph/conditionalaccesspolicy_test.go
@@ -1,6 +1,7 @@
 package msgraph_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


msgraph/models.go
```diff
@@ -730,6 +731,28 @@ func (c ConditionalAccessGuestsOrExternalUsers) MarshalJSON() ([]byte, error) {
...
+			return
+		}
+		switch *c.ExternalTenants.MembershipKind {
+		case ConditionalAccessExternalTenantsMembershipKindAll:
+			c.ExternalTenants.ODataType = utils.StringPtr(externalTenantsTypeAll)
+		case ConditionalAccessExternalTenantsMembershipKindEnumerated:
+			c.ExternalTenants.ODataType = utils.StringPtr(externalTenantsTypeEnumerated)
+		}
+	}
+	setExternalTenantsObjectType(guestOrExternalUsers.conditionalAccessGuestsOrExternalUsers)
```
manicminer, line 754: As there's no need for reusability here, could we simplify this by forgoing the constant declarations and move the `if`+`switch` blocks out into the parent method, dropping the anonymous func? The anonymous struct embeds a pointer to the original struct so this should work?

```suggestion
	if c.ExternalTenants != nil && c.ExternalTenants.MembershipKind != nil {
		switch *c.ExternalTenants.MembershipKind {
		case ConditionalAccessExternalTenantsMembershipKindAll:
			c.ExternalTenants.ODataType = utils.StringPtr("#microsoft.graph.conditionalAccessAllExternalTenants")
		case ConditionalAccessExternalTenantsMembershipKindEnumerated:
			c.ExternalTenants.ODataType = utils.StringPtr("#microsoft.graph.conditionalAccessEnumeratedExternalTenants")
		}
	}
```

msgraph/conditionalaccesspolicy_test.go
```diff
@@ -198,3 +199,78 @@ func testUser_Delete(t *testing.T, c *test.Test, user *msgraph.User) {
...
+func assertJsonMarshalEquals(t *testing.T, value interface{}, expected string) {
+	bytes, err := json.MarshalIndent(value, "", "  ")
+	if err != nil {
+		t.Fatalf("Marshalling failed with error %s", err)
+	}
+	actual := string(bytes)
+	if actual != expected {
+		t.Errorf("Expected marshalled json to equal %s but was %s", expected, actual)
+	}
+}
```
manicminer, line 212: This might be a useful function to move into the `internal/test` package?

manicminer (changes requested):
@agileknight Thanks again for working on this. This mostly LGTM, just a couple of refactoring suggestions below if you can take a look?

manicminer (approved):
@agileknight I hope you don't mind, in the interest of getting this in for a release today, I've committed my review suggestions and will tag this shortly. Appreciate the work on this!

philippdevtech:
@manicminer Thanks, I like the changes you made!

PR #268, opened 2024-01-08
Adding `ApplicationFilter` to Conditional Access Policy

alexwilcox9:
Adding application filter as per the API reference
https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessapplications?view=graph-rest-beta#properties

Let me know if you want any tests added/updated for this


manicminer (approved):
Thanks @alexwilcox9, this LGTM 👍

PR #271, opened 2024-01-22
Improve performance of GET requests

moio:
👋 hamilton maintainers!

I am opening this to ask whether it would make sense to include some new code, or patch existing code, in order to improve resource consumption of the `Get` method - particularly memory usage but also CPU.

I came to this problem as part of my day work at SUSE for Rancher, which uses hamilton to implement Entra ID (formerly Azure AD) integration. In some high-scale use tests we have users with principals in the thousands, groups in the millions and group assignments per principal in the hundreds - and enough traffic to need multiple (tens of) goroutines to periodically refresh all these objects in parallel. In extreme cases we have seen hundreds of MBs to GiBs worth of heap objects, which tend to cause out-of-memory situations.

Here is an example of a similar flame graph:
image

Of course our usage pattern isn't typical, and we are working at ways to change the way we interact with the library. Nevertheless according to my analysis `Get` (and code called by it) does quite a bit of needless computation:

 - OData `json.Unmarshal` calls are amplified by a factor of 2 due to the error handling in `OData.UnmarshalJSON`
 - `FromResponse` calls, which drive `OData.UnmarshalJSON` calls, are amplified by a factor of 2 by `performRequest`
 - `Get` needlessly re-unmarshal responses once after `FasterResponse` (again with the amplification factors above), and especially
 - `Get` un-marshals and subsequently re-marshals responses if they are paginated. This unmarshaling-marshaling extra round is repeated by the number of pages

This PR cuts down unmarshaling to one pass only (divided in sub-passes via `json.RawMessage`) and eliminates remarshaling altogether.

I could not find ways to achieve this without altering the `OData` struct and structs/methods which use it (`ConsistencyFailureFunc`, `ValidStatusFunc`, `GetHttpRequestInput`), therefore, I opted for a completely new code path in `fasterclient.go`. The approach can be changed if API changes are acceptable.

For now and for the sake of discussion, I only patched the one endpoint that really hurts my use case: `ListGroupMemberships`. Changes are minimal other than using the `FasterGet` method in place of `Get` - ideally this could be extended to all other (internal) usages of `Get`.

I am also contributing code to exercise `ListGroupMemberships` more thoroughly so that the effects of the patch are evident. Furthermore I added a utility method that will output CPU and memory usage during the test run.

My results are as follows:

|             |          main |    faster_get |
|-------------|--------------:|--------------:|
| Run 1       |        610 ms |        273 ms |
| Run 2       |        817 ms |        122 ms |
| Run 3       |        643 ms |        353 ms |
| **Average** |    **690 ms** |    **249 ms** |

**CPU load is ~36.14% of the original**

|             |          main | faster_get |
|------------:|--------------:|-----------:|
|       Run 1 |        21 MiB |      3 MiB |
|       Run 2 |        22 MiB |      3 MiB |
|       Run 3 |        23 MiB |      3 MiB |
| **Average** |    **22 MiB** |  **3 MiB** |

**Memory consumption is ~13.64% of the original**

Is there any interest to merge such an improvement? What would be needed to have a similar PR accepted?

Thanks in advance, I hope this helps

### DIFF
--- a/internal/test/metrics.go
+++ b/internal/test/metrics.go
@@ -1,0 +1,48 @@
+package test
+
+import (
+	"fmt"
+	"runtime"
+	"runtime/metrics"
+	"time"
+)
+
+// WithMetrics runs f and prints the total CPU time and max heap memory used during its execution
+func WithMetrics(f func()) {
+	cpuSample := make([]metrics.Sample, 1)
+	cpuSample[0].Name = "/cpu/classes/user:cpu-seconds"
+	memorySample := make([]metrics.Sample, 1)
+	memorySample[0].Name = "/memory/classes/heap/objects:bytes"
+	runtime.GC()
+
+	ticker := time.NewTicker(10 * time.Millisecond)
+	done := make(chan bool)
+	go func() {
+		maxMemory := uint64(0)
+		for {
+			select {
+			case <-done:
+				fmt.Printf("Max heap memory: %d MiB\n", maxMemory/1024/1024)
+				return
+			case _ = <-ticker.C:
+				metrics.Read(memorySample)
+				memory := memorySample[0].Value.Uint64()
+				if memory > maxMemory {
+					maxMemory = memory
+				}
+			}
+		}
+	}()
+
+	metrics.Read(cpuSample)
+	cpuSecondsBefore := cpuSample[0].Value.Float64()
+
+	f()
+
+	metrics.Read(cpuSample)
+	cpuSecondsAfter := cpuSample[0].Value.Float64()
+	fmt.Printf("Total CPU time: %.0f ms\n", (cpuSecondsAfter-cpuSecondsBefore)*1000)
+
+	ticker.Stop()
+	done <- true
+}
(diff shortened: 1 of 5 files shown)


moio:
@manicminer do you have any impressions to share about this work?

manicminer:
Hi @moio, my apologies for the delay in responding. I have unfortunately not had the bandwidth to take a proper look at these proposed improvements. I am not sure if you are still using this project at Rancher, but regardless I would encourage you to take a look at the new [Microsoft Graph SDK published by HashiCorp](https://github.com/hashicorp/go-azure-sdk/tree/main/microsoft-graph). Whilst that project is primarily intended for use in HashiCorp projects, I am sure that performance improvements would be welcomed there, wherever compatibility is maintained. Although you may find that implementing your own API client is the better way to go - and if this is the case you might find that project useful as a reference implementation, or to use some of the base packages for authentication etc.

I am about to deprecate/archive this project in favor of the HashiCorp MS Graph SDK, and I won't be merging any more PRs at this time. I'm therefore going to have to close this, again my apologies for the delay in responding and in hindsight it would have been better to let you know about this trajectory much earlier, I am sorry about that. Thank you again for spending the time to work on this :pray:

moio:
Thanks anyway.

FTR in case somebody else in future stumbles upon this, Rancher ended up keeping a forked version of this repo with this PR in as a short-term solution and migrating over to the Microsoft SDK as a long-term one.

👋

PR #273, opened 2024-02-19
Support for `attributeSets`

tiwood:
Adds support for [attributeSets](https://learn.microsoft.com/en-us/graph/api/resources/attributeset?view=graph-rest-1.0)

**Warning!** attributeSets cannot be DELETED or RENAMED, they are permanent. (see [here](https://learn.microsoft.com/en-us/entra/fundamentals/custom-security-attributes-add?tabs=ms-powershell#add-an-attribute-set))

Related:
https://github.com/hashicorp/terraform-provider-azuread/issues/913

### DIFF
--- a/internal/test/testing.go
+++ b/internal/test/testing.go
@@ -102,6 +102,7 @@ type Test struct {
 	ApplicationTemplatesClient                *msgraph.ApplicationTemplatesClient
 	ApplicationsClient                        *msgraph.ApplicationsClient
 	AppRoleAssignedToClient                   *msgraph.AppRoleAssignedToClient
+	AttributeSetsClient                       *msgraph.AttributeSetsClient
 	AuthenticationMethodsClient               *msgraph.AuthenticationMethodsClient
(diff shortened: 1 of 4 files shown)


manicminer (changes requested):
Hi @tiwood, thanks for submitting this PR! Given that attributeSets are immortal and have a quota of 500 per tenant, could you update the test to use a predictable name? When the test runs, it could create the attributeSet only if it's not present, and then proceed to update it afterwards. WDYT?

manicminer:
Hi @tiwood, just checking in to see if you're able to continue work on this PR? Thanks!

manicminer:
This project is being deprecated, and PRs are no longer being merged here.

Please refer to the new [Microsoft Graph SDK](https://github.com/hashicorp/go-azure-sdk/tree/main/microsoft-graph) published by HashiCorp, which inherits many of the same paradigms as this SDK, and which is now exclusively used in the Terraform AzureAD provider.

PR #278, opened 2024-03-09
Adding AccessPackageResourceRoleClient

Varantha:
Hey, 

Hope you don't mind a random PR. 

This aims to implement this method into hamilton:

https://learn.microsoft.com/en-us/graph/api/accesspackagecatalog-list-accesspackageresourceroles?view=graph-rest-beta&tabs=http

This will list all the possible roles available for selection for a specific resource in a specific access package catalog.

This is a requirement to be able to solve this issue on the azuread provider: 

https://github.com/hashicorp/terraform-provider-azuread/issues/1066

The API is pretty weird and if you omit any of the OData parameters then it won't work correctly so it's just the one, quite specific method.

Let me know if you need anything else from me on this 😃


manicminer (approved):
Thanks for the PR @Varantha, it is very welcomed!

Given the rather specific nature of this API's behavior, I think your implementation is fine. The relevant tests are passing and this looks good to merge. Thanks again!

PR #281, opened 2024-05-16
Feature - Add support for Custom Security Attributes

BrendanThompson:
This PR will add:
- `attribute sets`
- `custom security attribute definitions`

### DIFF
--- a/internal/test/random.go
+++ b/internal/test/random.go
@@ -13,9 +13,16 @@ func init() {
 
 // RandomString returns a random alphanumeric string useful for testing purposes.
 func RandomString() string {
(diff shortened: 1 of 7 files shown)


manicminer (changes requested):
Thanks @BrendanThompson! As discussed in private, for the attribute sets if we can name them predictably and only create when absent, this will avoid running the tenant-wide limit. For attributes, we should deactivate them at the end of the test as these don't contribute towards the limit. Other than that, this LGTM and should be good to go with those changes 👍

manicminer (approved):
Thanks for making the updates @BrendanThompson, this LGTM 🚀

manicminer:
Test failure is unrelated

PR #289, opened 2024-09-24
groups: add GetTransitiveMembers

antonjah:
This PR extends the project to include transitive listing of group members as `msgraph.User`. The addition of this method should not be confused with `ListTransitiveMembers` as it returns a `*[]string`, which forces the user to first list, and then get each individual member by ID.

More information about the `transitiveMembers` API can be found [here](https://learn.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0&tabs=http)

### DIFF
--- a/msgraph/groups.go
+++ b/msgraph/groups.go
@@ -358,6 +358,41 @@ func (c *GroupsClient) ListMembers(ctx context.Context, id string) (*[]string, i
 	return &ret, status, nil
 }
 
+// GetTransitiveMembers retrieves all nested members of the specified Group.
+// id is the object ID of the group.
+func (c *GroupsClient) GetTransitiveMembers(ctx context.Context, groupId string, query odata.Query) (*[]User, int, error) {
+	query.Expand = odata.Expand{Relationship: "*"}
+
+	resp, status, _, err := c.BaseClient.Get(ctx, GetHttpRequestInput{
+		ConsistencyFailureFunc: RetryOn404ConsistencyFailureFunc,
+		OData:                  query,
+		ValidStatusCodes:       []int{http.StatusOK},
+		Uri: Uri{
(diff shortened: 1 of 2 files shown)


manicminer (approved):
Thanks for this @antonjah! I'll happily merge this, but please note that I am going to deprecate this project very soon. We (HashiCorp) recently released a Microsoft Graph SDK that is now being used exclusively in the Terraform AzureAD provider instead of this SDK.

I encourage you to take a look at https://github.com/hashicorp/go-azure-sdk/tree/main/microsoft-graph and consider migrating your project. Many of the same paradigms have been implemented in the new SDK, with many improvements, and because the new SDK is auto-generated, it has massively more coverage.

For this use case, the new SDK has a `ListTransitiveMembers()` method in the `microsoft-graph/users/stable/transitivemember` package, which returns a `[]DirectoryObject` - which itself is an interface with automatic type discrimination, so you can retrieve users, groups, service principals etc in the same request.

antonjah:
> Thanks for this @antonjah! I'll happily merge this, but please note that I am going to deprecate this project very soon. We (HashiCorp) recently released a Microsoft Graph SDK that is now being used exclusively in the Terraform AzureAD provider instead of this SDK.
> 
> I encourage you to take a look at https://github.com/hashicorp/go-azure-sdk/tree/main/microsoft-graph and consider migrating your project. Many of the same paradigms have been implemented in the new SDK, with many improvements, and because the new SDK is auto-generated, it has massively more coverage.
> 
> For this use case, the new SDK has a `ListTransitiveMembers()` method in the `microsoft-graph/users/stable/transitivemember` package, which returns a `[]DirectoryObject` - which itself is an interface with automatic type discrimination, so you can retrieve users, groups, service principals etc in the same request.

I will definitely look into that! :-)

Again, thanks.

